PR #6, opened 2021-01-19
added plotPrevalence and plotTaxaPrevalance

FelixErnst:


R/plotPrevalence.R
```diff
@@ -0,0 +1,322 @@
+#' Plot prevalence information
+#'
```
microsud, line 2: I tested this function:
```
plotTaxaPrevalence(GlobalPatterns, rank = "Family",
                   detections = c(0, 0.0001, 0.001, 0.01, 0.1, 1))
plotTaxaPrevalence(GlobalPatterns, rank = "Genus",
                   detections = c(0, 0.0001, 0.001, 0.01, 0.1, 1))
plotTaxaPrevalence(GlobalPatterns, rank = "Species",
                   detections = c(0, 0.0001, 0.001, 0.01, 0.1, 1))
```
From family to species the plot is 


image

With `SprockettTHData`  species and family level shows similar plot as above. 
E.g. below Genus plot that works, but we can have a min.prevalence option, like in the `microbiome` package to avoid over-crowding with large regions of blue shaded taxa in the end of the figure. 
image

R/plotPrevalence.R
```diff
@@ -0,0 +1,322 @@
+#' Plot prevalence information
+#' 
+#' \code{plotPrevalence} and \code{plotTaxaPrevalence} visualize information on
+#' prevalence information.
```
microsud, line 4: "visualize prevalence information".

R/plotPrevalence.R
antagomir: Yes, this would be good. Also: is the color scale legend now in [0, 1] or in [0, 100] - If the unit (%) is given then I think also the numbers should be in [0, 100]. @tvborman could check this?

R/plotPrevalence.R
antagomir: Yes, better.

R/plotPrevalence.R
```diff
@@ -0,0 +1,322 @@
...
+                        unique(as.character(sort(as.numeric(ans$X)))))
+    }
+    ans$Y <- factor(ans$Y,lvls)
+    ans
+}
+
+#' @importFrom ggplot2 ggplot geom_raster geom_point geom_line
+#'   scale_fill_distiller scale_y_discrete scale_x_discrete scale_x_continuous
+#'   theme_classic
+.prevalence_plotter <- function(plot_data,
```
microsud, line 259: Another way of visualization that is common for ecological analysis of microbiota data is the Abundance-Occupancy relationship. This is basically, plotting Abundance (x-axis) versus Prevalence (Y-axis) mostly at ASV or Genus level and coloring points by a specific taxonomic rank. E.g. figure 1AB of this article [Stopnisek et al.](https://www.biorxiv.org/content/10.1101/727461v3.full). What do you think?

antagomir: I think it would be useful. We already have examples and code in the microbiome tutorial and packages (microbiome / microbiomeutils).

FelixErnst: I will have a look

R/plotPrevalence.R
FelixErnst: 6b0bcc2

R/plotPrevalence.R
FelixErnst: See option `layout = "point"` for `plotTaxaPrevalence`. (Also in examples)

FelixErnst: split of the plot type into extra function

antagomir (approved):

PR #9, opened 2021-02-09
Added facet functionality to plotPrevalentAbundance

tvborm:
Hi!

I added functionality to plotPrevalentAbundance. Now it is possible to create facets.

-Tuomas

### DIFF
--- a/R/plotPrevalence.R
+++ b/R/plotPrevalence.R
@@ -245,11 +245,13 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
                                                colour_by = colour_by,
                                                size_by = size_by,
                                                shape_by = shape_by,
-                                               label = label)
+                                               label = label,
+                                               facet_by = facet_by)
         plot_data <- vis_out$df
         colour_by <- vis_out$colour_by
         size_by <- vis_out$size_by
         shape_by <- vis_out$shape_by
+        facet_by <- vis_out$facet_by
         xlab <- paste0(ifelse(as_relative, "Rel. ", ""),"Abundance")
         ylab <- paste0("Prevalence(", ifelse(is.null(rank), "Features", rank),
                        ") [%]")
@@ -261,8 +263,7 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
                             size_by = size_by,
                             shape_by = shape_by,
                             ...)
-        
-        
+
         # If facet_by is not NULL, user has specified it. Adds the facets to the plot.
         if(!is.null(facet_by)){
             plot <- plot + 
@@ -274,7 +275,7 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
                       panel.background=element_blank(),
                 ) +
                 # Create facets
-                facet_wrap(~as.factor(rowData(x)[[facet_by]])) 
+                facet_wrap(~as.factor(rowData(x)[["Kingdom"]])) 
         }
         
         return(plot)
@@ -304,10 +305,12 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
                                         colour_by = NULL,
                                         size_by = NULL,
                                         shape_by = NULL,
-                                        label = NULL){
+                                        label = NULL,
+                                        facet_by = NULL){
     variables <- c(colour_by = colour_by,
                    size_by = size_by,
-                   shape_by = shape_by)
+                   shape_by = shape_by,
+                   facet_by = facet_by)
     if(!is.null(variables)){
         for(i in seq_along(variables)){
             # get data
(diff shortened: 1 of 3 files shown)


antagomir (approved):

R/plotPrevalence.R
```diff
@@ -38,6 +38,11 @@
 #'   \code{by} argument in 
 #'   \code{\link[scater:retrieveFeatureInfo]{?retrieveFeatureInfo}} for 
 #'   possible values. Only used with \code{layout = "point"}.
+#'   
+#' @param facet_by Specification of a feature to create facets.
+#' Argument can only be used in function getPrevalentAbundance. Value of argument
+#' must be in taxonomicRanks of \code{x}.
```
FelixErnst, line 44: That the arguments is only usable for getPrevalentAbundance should be visible from the man page. In addition I think you meant `plotPrevalentAbundace`

Shorter text: "taxonomic rank to facet the plot by. Value must be of \code{taxonomyRanks(x)}"

R/plotPrevalence.R
```diff
@@ -209,12 +220,19 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
              size_by = NULL,
              shape_by = NULL,
              label = NULL,
+             facet_by = FALSE,
```
FelixErnst, line 223: It would be easier to follow the other variables by setting the default to `NULL`

R/plotPrevalence.R
```diff
@@ -209,12 +220,19 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
...
         # input check
         .check_abund_values(abund_values, x)
         if(!.is_a_bool(as_relative)){
             stop("'as_relative' must be TRUE or FALSE.", call. = FALSE)
         }
+        # Check facet_by. It is FALSE by default, but user can specify it, but 
+        # the value must be in taxonomyRanks.
+        if(!(facet_by == FALSE || facet_by %in% taxonomyRanks(x))){
+            stop("'facet_by' must be in taxonomyRanks.",  call. = FALSE)
+        }
```
FelixErnst, line 234:
```
if(!is.null(facet_by)){
   ...
}
```

R/plotPrevalence.R
```diff
@@ -235,14 +253,31 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
...
+                theme(plot.background=element_blank(), 
+                      panel.grid.major=element_blank(),
+                      panel.grid.minor=element_blank(), 
+                      panel.background=element_blank(),
+                ) +
+                # Create facets
+                facet_wrap(~as.factor(rowData(x)[[facet_by]])) 
+        }
+        
+        return(plot)
```
FelixErnst, line 280: I think you should integrate this into `.incorporate_prevalence_vis` instead of bolting it on like this. If you choose to keep it like this, you should make sure, that the contract, that all data relevant for plotting is present in `object`, applies to this case as well.

FelixErnst (changes requested):

R/plotPrevalence.R
tvborm: yes, true

--> NULL

R/plotPrevalence.R
tvborm: Changed

R/plotPrevalence.R
tvborm: Changed

R/plotPrevalence.R
tvborm: Yep, that is better to have all the data in the object

I added facet_by also to .incorporate_prevalence_vis

R/plotPrevalence.R
```diff
@@ -274,7 +275,7 @@ setMethod("plotPrevalentAbundance", signature = c(x = "SummarizedExperiment"),
                       panel.background=element_blank(),
                 ) +
                 # Create facets
-                facet_wrap(~as.factor(rowData(x)[[facet_by]])) 
+                facet_wrap(~as.factor(rowData(x)[["Kingdom"]]))
```
FelixErnst, line 278: I think this should be `facet_wrap(vars(!!sym("facet_by")))`. Without testing I am not 100% sure, that the quosure stuff is correct. However, it is necessary since `facet_wrap(. ~ facet_by)` will/should get flagged by R CMD check due to the variable not being defined.

btw: the theme_call should not be necessary, since a theme is already set. I you need to tweak these settings, the functionality should be moved into `.prevalence_plotter`

tvborm: Ok, thanks. I forgot that "kingdom" there, because it wasn't working after those modifications that I made and I tried to solve why. But now it is working, because I used your code. 

Well, I don't think it is necessary to have those theme modifications there. Maybe best to have one kind of plot from one function, i don't know... I deleted those

FelixErnst:
Can you make sure it is up-to-data compared to master and bump the version? Thx

FelixErnst (approved):

tvborm:
It was already up to date compared to master. I made version bump

FelixErnst:
Can you also update the date? Probably best to remove it, because that was a comment in the recent submission of `microbiomeDataSets`

tvborm:
Ok, now I deleted version bump

FelixErnst:
I meant the date, but I will do it myself.

tvborm:
Ok, sorry, I misunderstood

PR #11, opened 2021-02-15
Plot core taxa time series

tvborm:
Hi!

I am facing some issues with this function, and that's why this is not progressing well... Do you have time to give some suggestions? Like, if the main idea of this is doable...

1. I didn't found help from scater. So, I don't know if the core of function is suitable.  Also, because of the data issues, I haven't had possibility to check if it works. 

2. I did not found any TSE data that includes time points. I used moving_pictures phyloseq and converted it to TSE. There is a problem with meltAssay: all colData's values are NULL for some reason.

3. Because of these fundamental issues, there are still lots of functionality that should be implemented. Like parameters and so on.

-Tuomas


microsud:
Hi @tvborm 
The `SilvermanAGutData()` in `microbiomeDataSets` is a longitudinal amplicon dataset from `invitro` bioreactors. Something that could be of interest. I was working with for a similar thing but with phyloseq and if you could share your email ID on slack, I can send you the code.

antagomir:
If the MovingPictures conversion fails due to shortcomings in our own conversion functions, this would be useful to investigate as a separate issue.

FelixErnst:
I would take a step back and would like to discuss the plot in a broader context.

1. Function name
From my point of view, the plot should work for any numeric value and also possibly factor values, if time points are classified more broadly. Since the y values are abundance values, the name `plotAbundance` would be most applicable. 
However, it is already used and I don't think, the function can survive additional arguments without confusing anyone (https://microbiome.github.io/miaViz/reference/plotAbundance.html). Therefore, I would suggest `plotSeries` as a function name. It is short, it is not too specific and whether it is a time series, a treatment series or whatever series shouldn't matter.

2. Which argument are required and how would they be named?
`object` for the `SE` is kind of a given. The series information should be plotted on the x axis, so I guess `x` should define the column of the `colData` to use. That we want to plot abundance values is also clear, but which ones need to be defined. I would use an argument `y` (alternatively `features`) for this and would select elements of `rownames(object)` with this. `abund_values` would define the assay to use and `rank` (default = NULL) would again be used to do on-the-fly agglomeration. A colour wouldn't need to be defined, since this is already selected via `y`. However, the linetype and the size could be defined by additional information from `rowData`.

So I think, if you incorporate these things into a draft, we would a first working version. Make sure, that you keep to the steps of:
- validation of the input
- assembling the data via the getters `retrieveFeatureInfo` and `retrieveCellInfo`
- plotting via an internal function analogous to the other plotting function (`.series_plotter` would be a suitable name)
and split each step into a separate function.

Let me know, what you think.

tvborm:
Ok, thank you. Now I see better the whole thing.

Another thing: I updated R to the latest version that came yesterday. It is 4.0.4, but microbiomeDatasets requires 4.1.0 as seen in the screenshot. What is that 4.1.0?

Screenshot from 2021-02-16 17-29-16

FelixErnst:
Because it is only available on the Bioconductor devel branch. This is the case for submission from October to April and not the case from April to October (roughly speaking), because Bioconductor is on a 6 month release schedule and R on a 12 month release schedule. If you want to read up on it : https://bioconductor.org/developers/package-guidelines/ https://bioconductor.org/developers/package-submission/ and look for release cycle.

If you want to install `microbiomeDataSets`, just install it from the release branch on GitHub like this:
`BiocManager::install("microbiome/microbiomeDataSets@release")`

tvborm:
Ok, it's working. Thanks for the information

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
+#' Plot Series
+#'
+#' Function that plots core taxa against time. #####################################
```
FelixErnst, line 3: Th has to be a complete sentence. Can you formulated it in a way that explains the general applicability for any series information?

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+#' @param x
+#' A \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#' object.
+#'
+#' @param abund_values
+#' A single character value for selecting the
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}}
+#'   to be plotted.
+#'
+#' @param x_axis
```
FelixErnst, line 14: Based on other function, lets just call it `x`

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+#'  A single character value for selecting the column from 
+#'  \code{\link[SummarizedExperiment:SummarizedExperiment-class]{ColData}} 
+#'  that will specify values of x-axis. 
+#'  
+#' @param x_axis
+#'  A single character value for selecting the taxa from 
+#'  \code{\link[SummarizedExperiment:SummarizedExperiment-class]{rownames}}. 
+#'  This parameter specifies taxa whose abundances will be plotted.
```
FelixErnst, line 22: ???

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+#'  
+#'  @param rank a single character defining a taxonomic rank, that is used to
+#'  agglomerate the data. Must be a value of \code{taxonomicRanks()} function.
```
FelixErnst, line 25: indentation is not correct

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+                   rank = NULL){
+              
+              ###################### Input check #######################
+              # Check abund_values
+              .check_abund_values(abund_values, x)
+              
+              # Check x_axis
+              if( !(x_axis %in% names(colData(x))) ){
+                  stop("'x_axis' must be a name of column of colData(x)", call. = FALSE)
+              }
```
FelixErnst, line 75: These checks are handled by `retrieveCellInfo`. you can skip it here

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+NULL
+
+#' @rdname plotSeries
+#' @export
+setGeneric("plotSeries", signature = c("x"),
+           function(x,
+                    abund_values = NULL,
+                    x_axis = NULL,
+                    y_axis = NULL,
+                    rank = NULL)
```
FelixErnst, line 55: I think you can remove the specific arguments except for `object`. (Use `object` instead of `x` so you avoid the clash and the argument names are the same as for the other plotting functions)

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+
+
+#' @rdname plotSeries
+#' @export
+setMethod("plotSeries", signature = c(x = "SummarizedExperiment"),
+          function(x,
+                   abund_values = NULL,
+                   x_axis = NULL,
+                   y_axis = NULL,
+                   rank = NULL){
```
FelixErnst, line 66: I can understand only some of the default values. For `abund_values` and `x` it does not make sense, where `rank` and `y` are fine

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+                  stop("'x_axis' must be a name of column of colData(x)", call. = FALSE)
+              }
+              
+              # Check y_axis
+              # If y_axis is not null, user has specified it
+              if (!is.null(y_axis)){
+                  if(!(y_axis %in% rownames(x)) ){
+                      stop("'y_axis' must be in rownames(x).", call. = FALSE)
+                  }
+              }
```
FelixErnst, line 83: I would include an else branch here and set `y <- rownames(object)`. This require the rank agglomeration having been done before this.

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+              # Check rank
+              # IF rank is not null, it has been specified by user
+              if( !is.null(rank) ){
+                  if(!.is_non_empty_string(rank)){
+                      stop("'rank' must be an non empty single character value,
+                           and it must be one of taxonomyRanks(x).",
+                           call. = FALSE)
+                  }
+                  .check_taxonomic_rank(rank, x)
+              }
```
FelixErnst, line 95: Here two if clause follow each other with the same check. Merge them and move the above the check for `y`. See comment above

R/plotSeries.R
```diff
@@ -0,0 +1,147 @@
...
+              color_by = "Kingdom"
+              line_by = "FeatureID"
+              
+              # # Take only those columns that are needed
+              # melted_data <- melted_data[, c("FeatureID", abund_values, x_axis)]
+              
+              ggplot2::ggplot(melted_data) + 
+                  ggplot2::geom_line(
+                      ggplot2::aes_string(x = x_axis, y = abund_values, 
+                                          color = color_by, linetype = line_by))
```
FelixErnst, line 131: Use the same function structure as used in the other plotting functions.

1. `.get_assay_data` should include retrieval of assay data. Include a check/warning for to much data. Maybe > 10 should get a warning and > 20 should get an error, since this will in every event not make sense. 
2. `.incorporate_vis` should include the incorporation of series information based on the colData, and colour, linetype, size parameter based on rowData. The additional columns should be named as the arguments, e.g. `X`, `colour_by`, `size_by`, `linetype_by`. `cbind` it to output of 1 and make sure to play back the found names, since partial name matching is supported by `retrieveCellInfo` and `retrieveFeatureInfo`.
3. `.melt_series_data` should include the melting step using `pivot_longer`. Here the series information has to be returned as `X` (capital letter) and the different assay values as `Y`. the rest of the additional information will get inflated alongside.
4. `.series_plotter` should be designed analogous to the other plotting functions including the set parameter for colour_by, size_by and linetype_by as well overloading via `...` for adjusting the default values for size, linetype showing legends, etc.

FelixErnst (changes requested):
Hi Tuomas,

thanks for the PR. Some work needs to be done.

Have a look at other plotting functions and how they are structured for inspiration. Let me know, if you have questions.

tvborm:
This version is still in progress and I have already added some internal functions. I am still working on with this, because I have been quite busy. Biggest problem I have had with implementing retrieveCellInfo and retrieveFeatureInfo, but I think these will help me to get all done.

R/plotSeries.R
tvborm: Now I have implemented .get_assay_data. 

But I have a problem with .incorporate_series_vis:

I get the feature_data, but where it should be stored? The feature_datas differs in length, so they can't be stored in same data frame just like that. 

Because the data is not melted, feature_datas can not be just assigned to xy-data points. 
I thought that I just take feature_datas as a list of lists, but I'm not sure if that is acceptable or smart thing to do.

FelixErnst: the feature data should have the same length as the chosen dimension. Beaware, that `retrieveCellInfo/FeatureInfo` return a named list: `value` is the actual data and `name` is the retrieved name. Due to the support for partial matching this is necessary.

For insparation look at other plotting functions in `miaViz`. For example in `plotGraph.R`:

```
 for(i in seq_along(variables)){
    # get data
    feature_info <- type_FUN(se, variables[i],
                             exprs_values = by_exprs_values)
    feature_info_name <- feature_info$name
    # mirror back variable name, if a partial match was used
    var_name <- names(variables)[i]
    assign(var_name, .get_new_var_name_value(get(var_name),
                                             feature_info$name))
    # rename columns by their usage
    feature_info$name <- var_name
    graph_data <- .add_graph_data_or_warn(feature_info, graph_data,
                                          type = "nodes",
                                          feature_info_name)
}
```

R/plotSeries.R
```diff
@@ -0,0 +1,182 @@
...
+                   size_by = size_by)
+    
+    # If variables there are variables
+    if(!is.null(variables)){
+        
+        # Loops through variables
+        for(i in seq_along(variables)){
+            
+            # If the variable is in colData
+            if( variables[i] %in% names(colData(se)) ){
```
FelixErnst, line 154: This check is done by `retrieveCellInfo` and therefore not necessary

R/plotSeries.R
tvborm: Yes, of course... I was in the wrong track

FelixErnst:
A little bit of contextual discussion: linetype and size setting make only sense, if the data is per feature, since the data must be orthogonal to the series information, which is plotted on the x-axis.

Is that correct or am I mistaken? I am not 100% sure, since I didn't try plotting series data yet.

R/plotSeries.R
```diff
@@ -0,0 +1,330 @@
...
+    
+    # resolve the colours
+    plot_out <- .resolve_plot_colours(plot_out,
+                                      plot_data$colour_by,
+                                      colour_by,
+                                      fill = TRUE)
+    plot_out <- .resolve_plot_colours(plot_out,
+                                      plot_data$colour_by,
+                                      colour_by,
+                                      fill = FALSE)
```
FelixErnst, line 320: This only need to be done once, since the fill argument is not applicable for geom_line, if I remember correctly. So probably best to keep just the `fill = FALSE` call.

tvborm:
> A little bit of contextual discussion: linetype and size setting make only sense, if the data is per feature, since the data must be orthogonal to the series information, which is plotted on the x-axis.
> 
> Is that correct or am I mistaken? I am not 100% sure, since I didn't try plotting series data yet.

So, y-axis have the abundances per-taxa, x-axis have e.g. time. So, `linetype_by = "Phylum" `makes sense to me, because then one line represent one taxa. But `linetype_by =  "SampleType"` does not make sense, because one line / one taxa can get different values for `"SampleType"`.

Actually, after those modifications `linetype_by` is not even working right now even with `linetype_by = "Phylum"`, so I have to fix that.

R/plotSeries.R
tvborm: Done only once, fill = FALSE left

FelixErnst:
That because you are getting the data from the `colData`. `Phylum` and such are stored on the `rowData`

R/plotSeries.R
```diff
@@ -301,9 +300,10 @@ setMethod("plotSeries", signature = c(object = "SummarizedExperiment"),
                                 linetype_by = linetype_by,
                                 size_by = NULL,
                                 alpha = line_alpha,
+                                linetype = line_type,
                                 size = line_size)
     
-    line_args$args$mapping$group <- sym("colour_by")
+    #line_args$args$mapping$group <- sym("colour_by")
```
tvborm, line 306: Now, when I commented row 306, it is working. I do not know, if grouping is necessary. I think it is done automatically.

FelixErnst: I am not sure. sometimes it, and sometimes it is not. I think you have to group based on the `series` identifier aka. the rowname, which the date to plot was derived from.

I think your are adding this to the `sample` column in `.melt_series_data`. It probably would be better named `feature` (Line 279 in the current state)

tvborm:
I checked that now melting is done correctly. There were no `feature`  column yet, that included taxa, so I added it. Also, now taxa is used for grouping.

I tried to fix he problem that occur sometimes when `linetype_by` is used simultaniously with `colour_by`. If colour is not constant through the line, it gives an error. I could not fix that, but I found out that it just might not be possible.

E.g.,
`x <- microbiomeDataSets::SilvermanAGutData()`

`plotSeries(x, abund_values = "counts", X = "DAY_ORDER", rank = "Phylum", colour_by = "SampleType", linetype_by = "Phylum")`

Causes error, because SampleType is changing, but
`
plotSeries(x[1:10], abund_values = "counts", X = "DAY_ORDER", colour_by = "Family", linetype_by = "Phylum")`

does not cause error, because both `Family` and `Phylum `are constant through lines

FelixErnst:
Hi Tuomas,

I saw that you are still working on this. You made good progress.


As a bit of input, I prepared a patch for your consideration to stream line some suggestion. We cannot go through the changes line by line and I didn't want to just push the changes, so lets try doing this via a patch.

Just apply the patch using `git apply file.patch`

[0001-refactoring-of-plotSeries-internals.zip](https://github.com/microbiome/miaViz/files/6077251/0001-refactoring-of-plotSeries-internals.zip)

... and go through the changes I would like to suggest. If things are unclear about the reason or how it works, let me know and we can discuss it.

Felix

tvborm:
Thank you! I checked the whole thing and tried to learn from it. This way was probably better because it was easier to get this finished with this kind of approach and I saw how it is done properly... And the plot looks pretty nice now.

One thing: should it be able to assign values to `colour_by` also from `colData`? I think now it is only trying to retrieve data from `rowData`, so currently `colour_by` can only get data from `rowData`.

FelixErnst:
I think you are approaching this too technical and should take a step back to a conceptual point of view.

I would propose the following questions to unriddle this:
- would a change in colour along time make sense for this plot type? (If yes, would this work technically, with geom_line?)
- Are there alternative plot types, which could highlight qualitative changes over time or any other series data? (colour works best with qualitative data, factors or character values. The hidden bit in your question is about quantiative versus qualitative, at least in part)
- What arguments would that plot type need?
- Would it be compatible with the arguments of `plotSeries`?
- Would you be able to reuse code you have written for `plotSeries`?

If you think along those lines, I think you can come up with a solution.

Let me know, if you want to add things to this PR or whether it should proceed with the current framing. Thanks.

Regarding the changes I suggested:
The approach I took is mostly guided by the same principle outlined above with these questions: How does data relate to each other in the context of one directional variables?
I think you should have a look at the other plotting functions and go through them. Have you ever used `debug` to step into a function and follow the functions progress step by step? If not, try it. It will make the next part you want to implement a bit easier. (hint: try `debug(miaViz:::.some_internal_function)`)

FelixErnst:
In addition you have to fix the tests

tvborm:
> would a change in colour along time make sense for this plot type? (If yes, would this work technically, with geom_line?)

- It might me beneficial to highlight e.g. if treatment is started. But I think that it wont work with geom_line.

> Are there alternative plot types, which could highlight qualitative changes over time or any other series data? (colour works best with qualitative data, factors or character values. The hidden bit in your question is about quantiative versus qualitative, at least in part)

- Histograms came to my mind. 

> 
> What arguments would that plot type need?
- e.g. time points, taxa data and data from colData. Time point would have e.g. 2 categories (e.g. treatment yes/no), and taxa would be recognized by different colours. 

> Would it be compatible with the arguments of plotSeries?

- Yes, only extra data from colData would be needed

> Would you be able to reuse code you have written for plotSeries?

- Yes.

Although, I think now the function is clear and does the thing that the user is expecting. Additional functionality would mess things, so I am  not sure if histogram is needed. 


-------------------------------------------------

I have not used debug before. I have just printed information, but this debug is very helpful, thanks! I will check those other functions also with debug. 

------------------------------------------------------

I fixed tests

FelixErnst:
> I fixed tests
 Thanks. Can you take care of the rounding error?

FelixErnst:
Regarding plotting additional data along the series data:

your points are valid, but I think you also realized, that a simpler first version can get you of the ground better, than going for the complicated full solution from the start.

Furthermore, please have a look at `plotAbundance`. I implemented similar things there, which might be resusable here. In short, if data from `colData` is added, an additional plot is generated and attached to the original series plot. This works with factor data and continous data (histrogram-like or an additional line plot).

I think it would be good to merge the PR now, when the tests are fixed and the build goes through. (Have a look at the GHA output for details). You can then add an additional PR, once you have a solution for adding the additional data plots. Would that work for you?

tvborm:
Ok, the plan sounds great. It's better to merge this PR now and do some additional things in other PR

PR #16, opened 2021-04-30
add plot*Tile function

FelixErnst:


R/plotColTile.R
```diff
@@ -0,0 +1,206 @@
...
+                                alpha = rect_alpha,
+                                colour = rect_colour)
+    rect_args$args$mapping$xmin <- sym("xmin")
+    rect_args$args$mapping$xmax <- sym("x")
+    rect_args$args$mapping$ymin <- sym("ymin")
+    rect_args$args$mapping$ymax <- sym("y")
+    # start plotting
+    plot_out <- ggplot(data) 
+    plot_out <- plot_out +
+        do.call(geom_rect,rect_args$args)
```
microsud, line 184: Should `NA` be given color? EDIT: I see the color now its is a shade of very light grey?
```
library(microbiomeDataSets)
library(miaViz)
tse <- SprockettTHData()
plotColTile(tse,"Age_Class","Community_Subtype")
```

R/plotColTile.R
```diff
@@ -0,0 +1,206 @@
...
+        .plot_tile_data(object, type = "column", x, y, 
+                        factor_x_by = factor_x_by, factor_y_by = factor_y_by,
+                        ...)
+    }
+)
+
+#' @rdname plotColTile
+#' @export
+setMethod("plotRowTile", signature = c("SummarizedExperiment"),
+    function(object, x, y, factor_x_by = NULL, factor_y_by = NULL, ...){
```
microsud, line 60: The `plotRowTile` will be tricky because of the high number of variables. 
```
plotRowTile(tse, x="Phylum", y="Family")

```
It is an important function to check which taxa within a given rank are dominant and will be useful. One option I think is to maybe structure it like giving top 3-5 within each Phylum to be plotted with an `other` option? It maybe difficult to achieve this but do you think it is possible?

R/plotColTile.R
```diff
@@ -0,0 +1,206 @@
...
+    plot_out <- plot_out +
+        do.call(geom_rect,rect_args$args)
+    # add scales
+    plot_out <- plot_out +
+        scale_x_continuous(name = paste0("Fraction (",xlab,")"),
+                           expand = c(0,0),
+                           breaks = seq(0,1,0.1),
+                           sec.axis = dup_axis(name = xlab,
+                                               breaks = coord$xmid,
+                                               labels = coord$X)) +
```
microsud, line 192: Here, for `plotRowTile` the sec.axis labels are crowded and maybe default can be 
`theme(axis.text.x.top = element_text(angle = 90, hjust = 0))` 
Edit: not sure if there is function to repel text on sec.axis?

R/plotColTile.R
```diff
@@ -0,0 +1,206 @@
...
+#' 
+#' @return 
+#' A \code{ggplot2} object or \code{plotly} object, if more than one 
+#' \code{prevalences} was defined.
+#' 
+#' @name plotColTile
+#' 
+#' @examples
+#' data(GlobalPatterns)
+#' se <- GlobalPatterns
```
microsud, line 33: Did you leave `plotRowTile(se, x="Phylum", y="Class")` out in example on purpose?

R/plotColTile.R
FelixErnst: I think this is the user's job. We cannot prevent `garbage in, garbage out`.

R/plotColTile.R
FelixErnst: Yes, but I think the user has to do that as well. If there are two many labels, the user will not and can add exactly that snippet on his own

R/plotColTile.R
FelixErnst: Yes. `plotRowTile` is less usable, since the data us usually longer. However, since it is technically possible and does not require any changes, I implemented it as well.

R/plotColTile.R
FelixErnst: Add a `na.value` setting to customize the color.

microsud (approved):
Good to go!

R/plotColTile.R
microsud: I thought so

FelixErnst:
> Good to go!

Can you merge?

Issue #15, opened 2021-04-17
New function: plot_density ?


antagomir:
Can we add a function for quick visualization of adundance profiles as a density plot?

The [microbiome pkg equivalent](https://github.com/microbiome/microbiome/blob/master/R/plot_density.R).

```
library(microbiome)
data(dietswap)
pseq.rel <- transform(dietswap, 'compositional')
library(ggplot2)
plot_density(pseq.rel, variable='Dialister') + scale_x_log10()
```

And after the basic functionality, it would be possible to add coloring schemes like in [this example](https://github.com/antagomir/pics-public/blob/master/TippingElements/Bimodal.png):

```
data(atlas1006)
pseq <- subset_samples(atlas1006, DNA_extraction_method == 'r')
pseq <- transform(pseq, 'compositional')
hotplot(pseq, 'Dialister', tipping.point=1.1, log10=TRUE)
```

FelixErnst:
I think this is a variation on the `plotExpression`/`plotAbundance` functions and can be easily added there. It just require a different plotter to be added (Are those ridge plots?).

The following example comes close I guess.

```r
se <- microbiomeDataSets::atlas1006()
se <- relAbundanceCounts(se)
altExps(se) <- splitByRanks(se)
plotExpression(altExp(se,"Genus"),
               exprs_values = "relabundance",
               features="Dialister",
               colour_by = "Dialister") + 
    scale_y_log10() + 
    coord_flip()
```

The additions required would be:
- new generic and method for plotting ridge plots
- argument for plotting log-transformed which also takes care of colouring

antagomir:
Yes, ridge plot. But can be also done just for a single taxonomic group.

FelixErnst:
Yep, but that is something easily extended to, if more than feature is plotted.

tvborm:
I wasn't completely sure what is the correct type of plot that is wanted. I thought that, when I have time I could take a closer look on this, because this issue is still open

So does this kind of plot look right? It utilize `ggridges` package

```
#install.packages("ggridges")
library(miaViz)
library(ggridges)

se <- microbiomeDataSets::atlas1006()

se <- transformSamples(se, method = "relabundance")

se_agg <- agglomerateByRank(se, "Genus")[c("Dialister", "Aeromonas", "Akkermansia", "Vibrio")]

#se_agg <- agglomerateByRank(se, "Genus")[c("Dialister")]

se_abund_df <- meltAssay(se_agg, assay_name = "relabundance")

[...]

Issue #13, opened 2021-04-17
New function: plot_spread ?


antagomir:
I suggest to add a new function that visualizes distribution of abundances per taxonomic group as a dotplot. We have used this in some publications and found it useful. Usefulness depends on the number of taxa, samples etc, as for any visualization.

Similar functionality has been implemented [in microbiome](https://github.com/microbiome/microbiome/blob/master/R/spreadplot.R) and works as follows:

```
library(microbiome)
data(dietswap)
spreadplot(transform(dietswap, "compositional"))
```

FelixErnst:
This might be already available via `scater`

```r
library(scater)
se <- microbiomeDataSets::dietswap()
altExps(se) <- splitByRanks(se)
plotHighestExprs(altExp(se,"Genus"))
```

![image](https://user-images.githubusercontent.com/17824332/115203059-56e2c300-a0f7-11eb-811d-ddb50a04e0ce.png)

Maybe a wrapper with a more microbiome specific name might be warranted, aka. `plotHighestAbundance`

microsud:
I was also thinking of the same function from scater and I have used it before for QC. IMO we need not add anything on top of it and just show this using scater in OMA?

antagomir:
This is a good plan. Although I have to admit that I like the dotplot a bit more compared to this one with vertical lines.

PR #17, opened 2021-06-11
Added plotAbundanceDensity

tvborm:
Hi,

here is the first version of `plotAbundanceDensity`. I tried to include here only the necessary things, because I thought it is easier to start improving from here.

Issues: https://github.com/microbiome/miaViz/issues/13 and https://github.com/microbiome/miaViz/issues/15

-Tuomas


R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,163 @@
...
+                                     colour_by = colour_by)
+        return(plot_out)
+    }
+)
+
+################################ HELP FUNCTIONS ################################
+
+.incorporate_density_data <- function(object, abund_values, n, colour_by){
+    # Gets the assay
+    abundance_table <- assay(object, abund_values, withDimnames=FALSE)
```
FelixErnst, line 101: Using the getter assay `withDimnames = FALSE` is a risk. The subsetting by name in line 105 isn't guaranteed to work in that case, since the assay dimnames maybe set to other values. Have a look at the man page for `assay`

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,163 @@
...
+    } else {
+        # Specifies aesthetic, when colour_by is not used
+        aesthetic <- aes_string(y = "taxa", x = "value")
+    }
+    
+    # Creates a list that includes the data and aesthetic
+    density_data_list <- list(density_data = density_data, 
+                              aesthetic = aesthetic)
+    return(density_data_list)
+}
```
FelixErnst, line 140: As mentioned above, this functions shouldn't deal with the aesthetics.
 In addition I think it can be shortened a bit
```r
# Gets the assay
    mat <- assay(object, abund_values)
    # Gets the most abundant taxa
    top_taxa <- getTopTaxa(object, n)
    # Subsets abundance table  by taking taxa of highest abundance
    mat <- mat[top_taxa, , drop=FALSE]
    # melt the data
    plot_data <- t(mat) %>%
        as.data.frame() %>%
        rownames_to_column("Sample") 
    # Gets coloring information if 'colour_by' is not NULL, otherwise specifies just aesthetic
    if (!is.null(colour_by)) {
        # Gets information from colData
        colour_out <- retrieveCellInfo(object, colour_by)
        # Mirrors back the variable name, if a partial match was used
        colour_by <- colour_out$name
        # Adds information to the data frame that includes all density data
        plot_data$colour_by <- colour_out$value
    }
    cols <- c("Sample","colour_by")[c("Sample","colour_by") %in% colnames(plot_data)]
    plot_data <- plot_data %>%
        pivot_longer(cols = !cols, names_to = "Y", values_to = "X")
    return(list(plot_data = plot_data, 
                colour_by = colour_by))
```

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,163 @@
...
+              axis.text.y = element_text(colour = text_colour),
+              axis.title.x = element_text(colour = text_colour),
+              axis.title.y = element_text(colour = text_colour),
+              title = element_text(colour = text_colour))
+    
+    if (!is.null(density_data$colour_by)) {
+        plot_out <- .resolve_plot_colours(plot_out, density_data$colour_by, colour_by)
+    }
+    return(plot_out)
+}
```
FelixErnst, line 163: Have a look at the other plotting functions and use the helper functions like `.resolve_plot_colours`, `.get_point_args`

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,163 @@
...
+                             aesthetic,
+                             abund_values,
+                             colour_by,
+                             point_alpha = 0.6,
+                             point_shape = 124,
+                             text_size = 8,
+                             text_colour = "gray35"){
+    # Creates the plot and annotations. 
+    plot_out <- ggplot(density_data, aesthetic) + geom_point(alpha = point_alpha, shape = point_shape)
+    plot_out <- plot_out + xlab(abund_values) + ylab("Feature")  + theme_bw(text_size) +
```
FelixErnst, line 152: Don't hard code a label. Make it an optional argument for `.density_plotter`

FelixErnst (changes requested):
Hi Tuomas @tvborm 

Thanks for the PR. It needs some restructuring and should use more of the common internal helper functions.

Have a look at the other plotting functions in this package for inspiration.

If the default is plotting points, then shape and size should also be adjustable, shouldn't they?

Felix

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,163 @@
...
+    function(object,
+             abund_values = "counts",
+             n = 50, 
+             colour_by = NULL, 
+             ...){
+        ############################# Input Check ##############################
+        # Checks abund_values
+        .check_assay_present(abund_values, object)
+        # Checks n
+        if( !(n%%1==0 && n>0) ){
```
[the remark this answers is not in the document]
tvborm: Fixed

R/plotAbundanceDensity.R
tvborm: Removed

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,163 @@
...
+    )
+    
+    # Gets coloring information if 'colour_by' is not NULL, otherwise specifies just aesthetic
+    if (!is.null(colour_by)) {
+        # Gets information from colData
+        colour_out <- retrieveCellInfo(object, colour_by)
+        # Mirrors back the variable name, if a partial match was used
+        colour_by <- colour_out$name
+        # Adds information to the data frame that includes all density data
+        density_data$colour_by <- colour_out$val[density_data$sample]
```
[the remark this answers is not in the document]
tvborm: Should have been more careful...

Changed

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,181 @@
...
+                                          fill = TRUE)
+    }
+    plot_out <- plot_out +
+        theme_classic()
+    # add legend
+    plot_out <- .add_legend(plot_out, add_legend)
+    # flip
+    plot_out <- .flip_plot(plot_out, flipped, add_x_text)
+    return(plot_out)
+}
```
FelixErnst, line 181: please add a newline at the end of the file

tvborm: Added

tvborm:
Now I have taken into account these things that were mentioned. I made small changes also to` utils_plotting.R,` because e.g. coloring did not work without them

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,171 @@
...
+        density_out <- .get_point_args(colour_by,
+                                       shape_by = NULL,
+                                       size_by = NULL,
+                                       alpha = point_alpha,
+                                       shape = point_shape,
+                                       size = point_size)
+        plot_out <- plot_out +
+            do.call(geom_point, density_out$args)
+        # resolve the colours
+        plot_out <- .resolve_plot_colours(plot_out,
```
FelixErnst, line 158: If `colour_by` is `NULL` this will return a warning.

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,171 @@
...
+                             add_legend = TRUE,
+                             flipped = FALSE,
+                             add_x_text = FALSE){
+    # start plotting
+    plot_out <- ggplot(density_data, aes_string(x="X", y="Y")) +
+        xlab(xlab) +
+        ylab(ylab)
+    
+    # Layout "density" or "point", "density" will be added
+    if (layout == "point"){
```
FelixErnst, line 148: If I set `layout = "whatever"` and empty plot will be returned. There must be hard fallback, probably as an `else`

FelixErnst: In addition, the layout is `point` by default, but it does not really return a point shape by default. This is misleading. Does anything speak against using a jitter type of plot as the default and switch to density as an alternative?

R/utils_plotting.R
```diff
@@ -253,15 +253,20 @@ NULL
...
         aes_args$shape <- "shape_by"
     }
     if (!is.null(colour_by)) {
-        aes_args$fill <- "colour_by"
+        # Only shapes 21 to 25 can be filled. Filling does not work in other shapes.
+        if(shape >= 21 && shape <= 25){
+            aes_args$fill <- "colour_by"
+        } else {
+            aes_args$colour <- "colour_by"
+        }
```
FelixErnst, line 269: If you code it like this you also have to update the value of `fill_colour`.

R/plotAbundanceDensity.R
antagomir: I think it is OK to put density as the default if this is easier.

FelixErnst:
Looks better. See the other comments. I think you should now finish this by adding all the options left on the table. Go for it!

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,171 @@
...
+#' @param colour_by a single character value defining a column from \code{colData}, that is used to
+#'   color plot. Must be a value of \code{colData()} function.
+#'   
+#' @param ... additional parameters for plotting. See 
+#'   \code{\link{mia-plot-args}} for more details
+#'
+#' @details
+#' This function plots abundance of the most abundant taxa. Abundance is plotted as
+#' a point plot, where x-axis represents abundance and y-axis taxa. Each point represents
+#' abundance of individual taxa in individual sample. Most common abundances are shown
```
FelixErnst, line 24: Make sure, that the man page picks up all the options and describes them clearly. Especially if option are disregarded because of other options (e.g. the layout option), it should be spelled out specifically. 

Here you have to add the `layout` option directly and not show them in the seperate `...` plot-args man page.

R/plotAbundanceDensity.R
tvborm: Fixed

R/utils_plotting.R
tvborm: Added

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,187 @@
...
+        density_out <- .get_point_args(colour_by,
+                                       shape_by = NULL,
+                                       size_by = NULL,
+                                       alpha = point_alpha,
+                                       shape = point_shape,
+                                       size = point_size,
+                                       position = "identity")
+        plot_out <- plot_out +
+            do.call(geom_point, density_out$args)
+    }
```
FelixErnst, line 158: Still this will produce an empyt plot, if `layout = "whatever"` is set

tvborm: Oh, sorry. My intention was to continue from here and add `else`, that would create a density plot

tvborm:
Now there is also `density plot` option, and information is added to man page

tvborm:
Currently, zeros are not handled if log-scaling is done to x-axis. Should those values be handled somehow (by adding small number?)

antagomir:
Add a small pseudocount. By default, min(x[x>0])/2

FelixErnst:
I wouldn't do that. The warning should be there and for a good reason. In my opinion a plotting function should not change the data.

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,259 @@
...
+                             log_scale_x_axis = FALSE){
+    # start plotting
+    # Density plot needs different kind of structure
+    if (layout == "density"){
+        # # Reorders the levels to reverse order so that the taxa with highest abundance
+        # is on the top
+        density_data$Y <- factor(density_data$Y , levels = rev(levels(density_data$Y)) )
+        plot_out <- ggplot(density_data, aes_string(x="X", colour = "Y")) +
+            xlab(xlab) +
+            ylab(ylab)
```
FelixErnst, line 194: I don't think you should colour by "Y". The colour can then not be changed by the user. I think you should keep it as `aes_string(x="X", colour = "Y")` and supplement a specific different aesthetic in the function call

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,259 @@
...
+    if (layout == "point"){
+        density_out <- .get_point_args(colour_by,
+                                       shape_by = NULL,
+                                       size_by = NULL,
+                                       alpha = point_alpha,
+                                       shape = point_shape,
+                                       size = point_size,
+                                       position = "identity")
+        plot_out <- plot_out +
+            do.call(geom_point, density_out$args)
```
FelixErnst, line 212: Add a different aesthetic here, if you need to

FelixErnst, line 212: Also you don't need to repeat the whole code block, if just one option changes. I think you should rearrange the if statement to make it shorter and more simple

R/utils_plotting.R
```diff
@@ -272,11 +278,12 @@ NULL
         geom_args$fill <- "grey70"
     }
     if (is.null(shape_by)) {
-        geom_args$shape <- 21
+        geom_args$shape <- shape
     }
     if (is.null(size_by)) {
         geom_args$size <- size
     }
+    geom_args$position <- position
```
FelixErnst, line 286: I think you can remove this from the internal helper again and just add the option in the required case. The logic of just adding a value can be done in the main internal function

R/plotAbundanceDensity.R
FelixErnst: also why not use `geom_jitter`?

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,259 @@
...
+                                       size_by = NULL,
+                                       alpha = point_alpha,
+                                       shape = point_shape,
+                                       size = point_size,
+                                       position = "jitter")
+        plot_out <- plot_out +
+            do.call(geom_point, density_out$args)
+    }
+    if (log_scale_x_axis){
+        plot_out <- plot_out + scale_x_log10()
```
FelixErnst, line 239: a log scaled x axis should not be added in here. The user can always add it outside, if he needs to by actually just adding one element to the plot. I think this is not worth running the argument through all the functions.

R/plotAbundanceDensity.R
```diff
@@ -0,0 +1,259 @@
...
+                                                       colour_by = colour_by)
+        # Extracts the density data and aesthetic from the list
+        density_data <- density_data_list$density_data
+        colour_by <- density_data_list$colour_by
+        
+        # Gets the plot from plotter
+        plot_out <- .density_plotter(density_data = density_data, 
+                                     layout = layout,
+                                     xlab = abund_values,
+                                     ylab = "Taxa",
```
FelixErnst, line 135: This is a hard coded value and it cannot be changed by the user. I think this should, if anything an argument with a default, but not hard coded.

R/plotAbundanceDensity.R
antagomir: also good

R/plotAbundanceDensity.R
antagomir: yes, jitter would be good

antagomir:
Right - I am good with Felix suggestion.

R/plotAbundanceDensity.R
tvborm: Changed

R/plotAbundanceDensity.R
tvborm: Removed

R/plotAbundanceDensity.R
tvborm: Changed

R/utils_plotting.R
tvborm: Removed

R/plotAbundanceDensity.R
tvborm: Added

R/plotAbundanceDensity.R
tvborm: I have been trying to do that way. That is why I chose to do density plots with facets, because I couldn't find a solution to arrange multiple density plots to the same picture

FelixErnst: I will have a look

FelixErnst: Have a look at the changes

tvborm: Now it looks very nice. `shape_by` and `size_by` wasn't working due to small bug, so I fixed them. Also I added some information to man page

FelixErnst:
I will wait until ggtree has the issue fixed on Bioc devel. The fix is available on GitHub, but not on the Bioc builder, yet.

FelixErnst (approved):

PR #30, opened 2021-07-30
Some updates

antagomir:
I guess these are ok, so merging directly if the test will pass. Just polishing.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,7 +21,7 @@ License: Artistic-2.0 | file LICENSE
 Encoding: UTF-8
 LazyData: false
 Depends:
-    R (>= 4.0),
+    R (>= 4.1.0),
     SummarizedExperiment,
     TreeSummarizedExperiment,
     mia (>= 0.99),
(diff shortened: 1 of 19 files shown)


DESCRIPTION
```diff
@@ -7,17 +7,22 @@ Authors@R:
...
-    includes plotting tree, graph and microbiome series data.
+    includes plotting tree, graph and microbiome series data. The package is part of the
+    broader miaverse framework.
 biocViews: Microbiome, Software, Visualization
 License: Artistic-2.0 | file LICENSE
 Encoding: UTF-8
 LazyData: false
 Depends:
-    R (>= 4.0),
+    R (>= 4.1.0),
```
FelixErnst, line 25: Please revert this to 4.0. This is not to be changed unless there is a functionality used introduced by R 4.1.

In addition it is against Bioc standard specify the minor version

R/plotPrevalence.R
```diff
@@ -17,7 +17,7 @@
 #'     \code{\link[=agglomerate-methods]{?agglomerateByRank}} for more details.
 #'   }
 #'   \item{additional arguments for plotting. See 
-#'   \code{\link{mia-plot-args}} for more details}
+#'   \code{\link{mia-plot-args}} for more details i.e. call \code{help("mia-plot-args")}}
```
FelixErnst, line 20: Why? These is a link next to it on the man page. Is this really required?

DESCRIPTION
antagomir: Right, I was mistakingly syncing this with mia version requirement on the same go. Reverting in a moment.

Thanks for the Bioc standard note, I had overlooked this.

PR #25, opened 2021-07-21
plotRowTree layout bugfix

TuomasBorman:
Hi,

here is bugfix for `plotRowTree`. 

Problem: `layout` was not passed all the way to the function that takes care of the edge coloring; `.plot_tree_edges`. Because of that the function used default layout to plot those edges --> colored, rectangular tree was plotted on top of another tree with intended layout. 

Solution: `layout` is passed to the `.plot_tree_edges`. `layout` is added as one of the arguments with `.get_edge_args `function. 

Issue: https://github.com/microbiome/miaViz/issues/23

-Tuomas

PR #31, opened 2021-07-30
added order_descending to plotAbundanceDensity

FelixErnst:


R/plotAbundanceDensity.R
```diff
@@ -3,33 +3,38 @@
...
 #' @param abund_values a single character value for selecting the
-#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
-#' plotted. (default: \code{abund_values = "counts"})
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{assay}} to be
+#'   plotted. (default: \code{abund_values = "counts"})
 #'
-#' @param n a positive integer specifying the number of the most abundant taxa to show.
-#' (default: \code{n = 25})
+#' @param n a positive integer specifying the number of the most abundant taxa
+#'   to show. (default: \code{n = 25})
```
antagomir: Let us make the default n=min(nrow(object), 25)?

antagomir (commented):
Thanks. 

I have not found out how to adjust the point coloring; they have now a black outer circle, and gray color by default. The black edge on the gray dots compromises figure interpretation when the number of dots is large; what I have in mind is pure gray dots with some alpha, as in a recent paper [Supplementary Fig S1](https://static-content.springer.com/esm/art%3A10.1038%2Fs41467-021-22962-y/MediaObjects/41467_2021_22962_MOESM1_ESM.docx)

antagomir (approved):

R/plotAbundanceDensity.R
```diff
@@ -108,82 +113,99 @@
...
+              # Checks size_by
+              if( !is.null(size_by) && !.is_a_string(size_by)){
+                  stop("'size_by' must be a single character value.",
+                       call. = FALSE)
+              }
+              # Checks size_by
+              if( !is.na(order_descending) && !.is_a_bool(order_descending)){
+                  stop("'order_descending' must be TRUE, FALSE or NA.",
+                       call. = FALSE)
+              }
```
TuomasBorman, line 174: Minor thing: comment is wrong

TuomasBorman (approved):

TuomasBorman:
> I have not found out how to adjust the point coloring; they have now a black outer circle, and gray color by default.


Is the problem that `colour` match with` colour_by`? Otherwise I think it should work, and `colour` should be passed further? Is there an argument that makes `colour_by` specific?

`Error in retrieveCellInfo(object, colour_by) : cannot find 'red' `

FelixErnst:
How about this:

```r
plotAbundanceDensity(tse[1:3,], abund_values = "relabundance",
                     order_descending = NA, point_shape = 19)
```

R/utils_plotting.R
```diff
@@ -349,9 +367,6 @@ NULL
     if (is.null(edge_size_by)) {
```
TuomasBorman, line 367: Does this break `plotTree`? Commit: https://github.com/microbiome/miaViz/commit/6bd92a8e9565749db90dad43f76a7c54043101f5

FelixErnst: Why should it?

TuomasBorman: ..that `layout` is removed. The comment is in wrong place

antagomir:
> ```r
> order_descending = NA, point_shape = 19)
> ```

Clearly better, yes, but edges in circles still remain. Passing the colour argument does not do the trick, and I have not yet found a solution by manipulating the points afterwards with `geom_point()` for instance.

R/utils_plotting.R
TuomasBorman: The problem was that edges were plotted with default layout. --> two different trees were plotted. PR: https://github.com/microbiome/miaViz/pull/25

FelixErnst: Keep it in mind, if you see a problem. I am not sure, why added it in the first place, but there must have been a reason. Maybe you should look into again with this mind. I don't remember the original issue with the two trees, but I think it worth investigating.

FelixErnst:
I think this is clearly a case of `ggplot2` huge functionality. I can only change the default colour with a bit of weird code, since `scale_fill_manual` works only, when there is a scale.

```r
p <- plotAbundanceDensity(tse[1:3,], abund_values = "relabundance",
order_descending = NA, point_shape = 19)
p$layers[[1]]$aes_params$fill <- "red"
p
```

The option advanced argument `point_colour` is now supported. However, I must say, that the default "grey70" looks better with the alpha set to `1`

```r
plotAbundanceDensity(tse[1:3,], abund_values = "relabundance",
                     order_descending = NA, add_legend = TRUE, point_colour = "black")
plotAbundanceDensity(tse[1:3,], abund_values = "relabundance",
                     order_descending = NA, add_legend = TRUE, point_colour = "black", point_alpha = 1) # Most horrible
plotAbundanceDensity(tse[1:3,], abund_values = "relabundance",
                     order_descending = NA, add_legend = TRUE, point_alpha = 1) # best
```

antagomir:
Hmm yes. If there is no easy way to manipulate the figure either through function arguments, or through post-hoc manipulation of the ggplot object, then I think this is a clear limitation for the current system logic. But if there is no solution available, this is still acceptable I think.

R/utils_plotting.R
TuomasBorman: **This was the problem**
git checkout 5cace72fb92072fd79375a6b8b2aacbb063ad8d2
```
data("GlobalPatterns")
se <- GlobalPatterns 
se <- agglomerateByRank(se, rank = "Phylum")
plotRowTree(se,
            edge_colour_by = "Kingdom", layout="ellipse")
```
  
image

 
**Solved in master:**
image

 
**order_descending:**
```
 Error in .get_edge_args(edge_colour_by, edge_size_by, alpha = line_alpha,  : 
  unused argument (layout = layout) 
```

FelixErnst: So the layout did not propagate or, was it not used by the `ggtree` function?

TuomasBorman: plotRowTree --> .plot_row_column_tree --> .tree_plotter --> .plot_tree_edges:

gets layout inside plot_out argument and as layout argument. 

The problem was that `layout` inside `plot_out` did not work here were edges are plotted. This part needed its own `layout` to work
image

FelixErnst:
```
 Error in .get_edge_args(edge_colour_by, edge_size_by, alpha = line_alpha,  : 
  unused argument (layout = layout) 
```

This should have errored already for you, since the you removed the layout argument from the function definition, but not from the call in line 1244.

So I fixed it by adding the layout argument again, but no using it.

However, jsut to get back to the original question: `geom_tree` with using the layout argument produced a different tree, than was already set in plot_out, correct? So I think this would be a bug in `ggtree`. Do you want to report it?

TuomasBorman:
```
library("mia")
data("GlobalPatterns")
se <- GlobalPatterns 

se <- agglomerateByRank(se, rank = "Phylum", agglomerateTree = TRUE)
tree <- rowTree(se)

plot_out <- ggtree(tree,
                   layout = "ellipse")

plot_out + geom_tree()
```

image

Looks like it, yes I can report it

FelixErnst: I think you have to simplify the example more. Just save the tree to file and upload as zip to the issue

Issue #44, opened 2022-01-22
plotAbundance error?


antagomir:
The `plotAbundance` method provides the `features` argument. 

This is intended to highlight colData variables below the plot. According to manpage:

> "features: a single ‘character’ value defining a column from ‘colData’
>           to be plotted below the abundance plot. Continuous numeric
>           values will be plotted as point, whereas factors and
>           character will be plotted as colour-code bar. (default:
>           ‘features = NULL’)"
> 

The following commands output ggplot object p, which is not a single figure with colData highlighted below the figure. Instead, this outputs two different figures. My impression is that either the function or the manpage has some mistake. The expected behaviour (to me) would be something like that explained on the manpage.

```
data(GlobalPatterns)
se <- GlobalPatterns
p <- plotAbundance(se, abund_values = "counts", rank = "Phylum", features = "SampleType")
names(p)
```

[1] "abundance"  "SampleType"

On the same go, I would like to see an example in manpages showing how one can add sampleIDs below the plot. This would be a common use case for this.

antagomir:
I suggest that @YagmurSimsekk could check this, with support from @TuomasBorman

FelixErnst:
TL;DR: This works as intended. I won't accept the PR, since it forces the combination of plots by default.

The man page very very clearly states:
```
Value
a ggplot object or list of two ggplot objects, if features are added to the plot.
```
So the function and the man page are both correct. 

I added an additional example for producing the result, I think you are looking for:
```r
library(miaViz)
library(patchwork)
data(GlobalPatterns)
se <- GlobalPatterns
plot <- plotAbundance(se, abund_values = "counts", rank = "Phylum", features = "SampleType")
plot[[1]] <- plotAbundance(se, abund_values = "counts", rank = "Phylum",
                           features = "SampleType", add_legend = FALSE,
                           add_x_text = TRUE)[[1]] +
                           theme(axis.text.x = element_text(angle = 90)) 
wrap_plots(plot, ncol = 1, heights = c(0.8,0.2))
``` 
Does this fit, what you are looking for?

antagomir:
Yes, this solution works as well, good examples are always valuable.

PR #45, opened 2022-01-26
Plot abundance

YagmurSimsekk:
Now the list is turned into single plot.


R/plotAbundance.R
```diff
@@ -129,6 +133,7 @@ setGeneric("plotAbundance", signature = c("x"),
 #' @rdname plotAbundance
 #' @importFrom scater plotExpression
 #' @importFrom ggplot2 facet_wrap
+#' @importFrom gridExtra grid.arrange
```
antagomir, line 136: We recommend the **patchwork** pkg for multi-panel plots. Could you change this?

antagomir (commented):
Thanks. And this is for issue #44

R/plotAbundance.R
```diff
@@ -18,92 +18,96 @@
...
-#'   
-#' @param order_rank_by How to order abundance value: By name (\dQuote{name}) 
+#'
+#' @param order_rank_by How to order abundance value: By name (\dQuote{name})
 #' for sorting the taxonomic labels alphabetically, by abundance (\dQuote{abund}) to
 #' sort by abundance values or by a reverse order of abundance values (\dQuote{revabund}).
-#'  
-#'   
+#'
+#'
```
TuomasBorman, line 32: Additonal lines could be deleted, here is two lines and others have one between @params

TuomasBorman (commented):
Hi,

looks good and solid. 

However, I am still thinking if we should only update manuals

1. This creates a new dependency
2. For example, if user wants to change colors of colData-plot (in the example case: if user wants to specify colors of sample types "Ocean", "Tongue", and so on), is it too complictaed when returned plot is patchwork, not ggplot?
3.  patchwork(plot) is relatively simple task
4. User might prefer some other tools like gridExtra

-Tuomas

antagomir:
I agree with @TuomasBorman but now investigated this in more detail.

The very first implementation by @FelixErnst (see [commit history](https://github.com/microbiome/miaViz/commits/master/R/plotAbundance.R) seems to include a case for combining plots, with `@importFrom cowplot plot_grid`. Hence, it is intended that plots are combined internally. We can keep the original version and dependency on that, if there is no reason to change. 

The problem was reported in issue #44 and the problem is that the function does not work as expected in a certain case as it returns two plots that are plotted separately, and not a single combined plot for the sample labels are shown below the main figure (in contrast to what the documentation says). 

Can we update this PR so that it solves the problem that is described in issue #44 so that the function will return, for that example, the intended type of plot as explained on the manpage? If this appears to be challenging, we can discuss more online.

YagmurSimsekk:
It has been updated to match what is explained in the man page. PR can be merged.

FelixErnst:
Doesn't solve the issue and forces combination of plot reducing massively the flexibility of the function

antagomir:
Yes. The best solution could be to open an entirely new PR that addresses the issue described in #44

FelixErnst:
The issues in #44 are addressed and solved. Nothing to do anymore or did I overlook something?

PR #71, opened 2023-03-20
Build fix

TuomasBorman:
miaTime cannot be found from Bioc nor CRAN --> since it is listed in IMPORTS miaTime is searched from them --> error

http://bioconductor.org/checkResults/devel/bioc-LATEST/miaViz/nebbiolo1-install.html


antagomir:
Thansk!

PR #75, opened 2023-04-03
Build fix

TuomasBorman:
miaTime cannot even be in "suggests" as it causes an error in build https://master.bioconductor.org/checkResults/3.17/bioc-LATEST/miaViz/nebbiolo1-checksrc.html

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,7 +47,6 @@ Imports:
     ape,
     DirichletMultinomial
 Suggests:
-    miaTime,
     knitr,
     rmarkdown,
     BiocStyle,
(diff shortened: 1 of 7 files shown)


antagomir:
> miaTime cannot even be in "suggests" as it causes an error in build https://master.bioconductor.org/checkResults/3.17/bioc-LATEST/miaViz/nebbiolo1-checksrc.html

Right: we should just finalize that package, and submit to Bioc, when time will allow..

antagomir:
Ok good to open issues in miaTime but seems you just opened some already.

antagomir (approved):

PR #80, opened 2023-06-28
Updated plotDMN to work with newest DMM version

BananaCancer:
With the recent/upcoming modifications to DMM in mia and the addition of the algorithm to bluster, this function needs to change to work with the newest version of DMM.
I changed it so that it still works as before by putting a warning if the DMM info is in the wrong place.
It works as intended if the tse contains the info in the old place (`metadata$DMM`) or new place (`metadata$DMM$dmm`)

### DIFF
--- a/.github/workflows/check-bioc-devel.yml
+++ b/.github/workflows/check-bioc-devel.yml
@@ -166,6 +166,7 @@ jobs:
           
           BiocManager::install("HDF5Array")
           BiocManager::install("fionarhuang/TreeSummarizedExperiment")
+          devtools::install_github("LTLA/bluster")
           BiocManager::install("microbiome/mia")
           devtools::install_github("microbiome/miaTime")	  
           
--- a/.github/workflows/check-bioc-devel.yml
+++ b/.github/workflows/check-bioc-devel.yml
@@ -168,7 +168,6 @@ jobs:
           BiocManager::install("fionarhuang/TreeSummarizedExperiment")
           BiocManager::install("microbiome/mia")
           devtools::install_github("microbiome/miaTime")
-          remotes::install_github("LTLA/bluster")
           
           # rspm is out of date
           install.packages("matrixStats", repos = "https://cran.rstudio.com/")
--- a/.github/workflows/check-bioc-devel.yml
+++ b/.github/workflows/check-bioc-devel.yml
@@ -167,7 +167,8 @@ jobs:
           BiocManager::install("HDF5Array")
           BiocManager::install("fionarhuang/TreeSummarizedExperiment")
           BiocManager::install("microbiome/mia")
-          devtools::install_github("microbiome/miaTime")	  
+          devtools::install_github("microbiome/miaTime")
+          remotes::install_github("LTLA/bluster")
           
           # rspm is out of date
           install.packages("matrixStats", repos = "https://cran.rstudio.com/")
--- a/.github/workflows/check-bioc-devel.yml
+++ b/.github/workflows/check-bioc-devel.yml
@@ -166,7 +166,6 @@ jobs:
           
           BiocManager::install("HDF5Array")
           BiocManager::install("fionarhuang/TreeSummarizedExperiment")
-          devtools::install_github("LTLA/bluster")
           BiocManager::install("microbiome/mia")
           devtools::install_github("microbiome/miaTime")	  
           
--- a/R/plotDMN.R
+++ b/R/plotDMN.R
@@ -23,14 +23,13 @@
 #' @name plotDMN
 #'
 #' @examples
-#' library("bluster")
 #' data(dmn_se, package = "mia")
-#' tse_dmm <- cluster(dmn_se, name = "DMM",
-#'                    DmmParam(k = 1:4, type = "laplace"),
-#'                    MARGIN = "samples", full = TRUE)
-#'
+#' 
+#' # Put metadata in correct place
+#' metadata(dmn_se)$DMM$dmm <- metadata(dmn_se)$DMN
+#' 
 #' # plot the fit
-#' plotDMNFit(tse_dmm, name = "DMM", type = "laplace")
+#' plotDMNFit(dmn_se, name = "DMM", type = "laplace")
 #' 
 NULL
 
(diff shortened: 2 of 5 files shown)


BananaCancer:
@antagomir @TuomasBorman Is this ok for you? I also added bluster in the workflow script and added bluster in suggests in the DESCRIPTION file. The NEWS and version have also been updated.

antagomir:
Seems good to me. Also check if this is something to show in OMA.

antagomir (approved):

.github/workflows/check-bioc-devel.yml
```diff
@@ -166,6 +166,7 @@ jobs:
           
           BiocManager::install("HDF5Array")
           BiocManager::install("fionarhuang/TreeSummarizedExperiment")
+          devtools::install_github("LTLA/bluster")
```
TuomasBorman, line 169: Does  BiocManager::install work for bluster? All other packages are installed from Bioc devel which should include these changes made in bluster

R/plotDMN.R
```diff
@@ -23,11 +23,15 @@
 #' @name plotDMN
 #'
 #' @examples
+#' library("bluster")
 #' data(dmn_se, package = "mia")
-#' names(metadata(dmn_se))
+#' tse_dmm <- cluster(dmn_se, name = "DMM",
+#'                    DmmParam(k = 1:4, type = "laplace"),
+#'                    MARGIN = "samples", full = TRUE)
```
TuomasBorman, line 30: How long this takes? `dmn_se` object already has DMM calculated --> you could just plot it without actually calculating it (This removes bluster also from suggests)

R/plotDMN.R
```diff
@@ -43,7 +47,13 @@ setGeneric("plotDMNFit", signature = "x",
 setMethod("plotDMNFit", signature = c(x = "SummarizedExperiment"),
     function(x, name = "DMN", type = c("laplace","AIC","BIC")){
         #
-        dmn <- getDMN(x, name)
+        if (!is.null(metadata(x)[[name]]$dmm)) {
+            dmn <- metadata(x)[[name]]$dmm
+        } else {
+            .Deprecated(old="getDMN", new="cluster", 
+                    "Now runDMN and calculate DMN are deprecated. Use cluster with DMMParam parameter and full parameter set as true instead.")
+            dmn <- metadata(x)$name
```
TuomasBorman, line 55: calculate DMN --> calculateDMn

Line 55 does not work similarly than it used to work --> metadata(x)[[name]]

.github/workflows/check-bioc-devel.yml
TuomasBorman: IF not, add comment which tells when this line could be changed to Bioc installation

TuomasBorman (changes requested):
good!

R/plotDMN.R
antagomir: Good point. On the other hand I think we could show how one can cluster some new data with DMM because it is such a common and customized method for microbiome sample clustering? The example data set could be something else than dmn_se and small data set might be preferred to have fast calculation.

BananaCancer: The issue is that `dmn_se` has data in `metadata(tse)$dmm` instead of `metadata(tse)$DMM$dmm`. This creates a deprecated warning and the check fails.
I will change the data to put dmm in the correct place

R/plotDMN.R
BananaCancer: True, my bad

antagomir:
We could also consider if the DMM slot there needs to be readily calculated. I don't think this is used very much in our examples?

BananaCancer:
> We could also consider if the DMM slot there needs to be readily calculated. I don't think this is used very much in our examples?

So for example, if neither metadata(tse)$DMM$dmm and metadata(tse)$DMM exist, we execute a basic DMM call.
So this example,
```r
tse <- plotDMNFit(tse, name = "DMM", type = "laplace")
```
would execute
```r
        if (!is.null(metadata(x)[[name]]$dmm)) {
            dmn <- metadata(x)[[name]]$dmm
        } else if (!is.null(metadata(x)[[name]]) {
            .Deprecated(old="getDMN", new="cluster", 
                    "Now runDMN and calculateDMN are deprecated. Use cluster with DMMParam parameter and full parameter set as true instead.")
            dmn <- metadata(x)[[name]]
        } else {
            tse <- cluster(tse, name = name, DmmParam(k = 1:3, type = "laplace")
            dmn <- metadata(x)[[name]]$dmm
        }
        # ...
        tse
```

R/plotDMN.R
BananaCancer: Sorry @antagomir I didn't see you reply, do you want me to change the example even if it adds bluster dependency? Or do I leave it as is?

antagomir: Do you mean it adds bluster dependency in OMA?

antagomir:
I was wondering what is the point of having data object that already includes something that users usually need to calculate themselves anyway?

R/plotDMN.R
BananaCancer: Using bluster here adds a dependency in miaViz.
We already have a better, more complete example in OMA that uses bluster and miaViz.
So the example in miaViz should stay fairly simple.

BananaCancer:
Oh ok, yeah I think it's pretty rare that it's the case. 
I can add the bluster dependency and change the example to start with a simple SE, add the DMM, and plot the fit.

TuomasBorman:
> I was wondering what is the point of having data object that already includes something that users usually need to calculate themselves anyway?

DMM can take a long time to calculate, that was the only reason

antagomir:
> > I was wondering what is the point of having data object that already includes something that users usually need to calculate themselves anyway?
> 
> DMM can take a long time to calculate, that was the only reason

Now it seems there was also a second reason: to avoid bluster dependency in miaViz

antagomir:
I am not sure if it really is a problem to have bluster as a dependency in miaViz. The example might be more clear if the clustering was also done as part of it. If the example uses a small data set with a small maximum number of clusters then perhaps the calculation does not take too long.

TuomasBorman:
> I am not sure if it really is a problem to have bluster as a dependency in miaViz. The example might be more clear if the clustering was also done as part of it. If the example uses a small data set with a small maximum number of clusters then perhaps the calculation does not take too long.

Yep, the dependency issue is not big --> but why have bluster as dependency when we already have mia as dependency (which imports bluster)

BananaCancer:
After testing, the check fails if bluster isn't in the `Suggests`, so I'll leave it there.

Is this ok to merge?

antagomir:
ok!

antagomir:
Merge if ready.

.github/workflows/check-bioc-devel.yml
```diff
@@ -167,7 +167,8 @@ jobs:
           BiocManager::install("HDF5Array")
           BiocManager::install("fionarhuang/TreeSummarizedExperiment")
           BiocManager::install("microbiome/mia")
-          devtools::install_github("microbiome/miaTime")	  
+          devtools::install_github("microbiome/miaTime")
+          remotes::install_github("LTLA/bluster")
```
TuomasBorman, line 171: This should be useless line after updates in mia is in Bioc devel

TuomasBorman (approved):

PR #87, opened 2023-08-25
plotRDA

TuomasBorman:
First version of `plotRDA` method. Any thoughts?

```
data(enterotype)
tse <- enterotype

tse <- tse[, complete.cases(colData(tse))]
tse <- runRDA(tse, data ~ Nationality + Gender, na.action = na.omit)

plotRDA(tse, dimred = "RDA", point_size = 4, colour_by = "Gender")
```


TuomasBorman:
@RiboRings Sorry, I just realized that I accidentally overwrote your branch... --> If you have time, you could develop this further (update document and checks and make the code better...)

antagomir:
Seems promising to me at least.

RiboRings:
Ok I'm playing with it

RiboRings:
I'm wondering why `plotReducedDim` only shows a limited number of samples, even when there is no missing data. And when I look into the `reducedDim` object all samples have coordinates and therefore they should also appear in the plot. For example:

```
library(mia)
library(miaViz)

data("Tengeler2020", package = "mia")
tse <- Tengeler2020

dim(tse)
[1] 151  27

any(is.na(tse$patient_status))
[1] FALSE
any(is.na(tse$cohort))
[1] FALSE

tse <- runRDA(tse,
              assay.type = "relabundance",
              formula = assay ~ patient_status + cohort,
              distance = "bray",
              na.action = na.exclude)

reducedDim(tse, "RDA")

          dbRDA1       dbRDA2      dbRDA3
A110 -0.17355472 -0.259042261 -0.14200871
A12  -0.17355472 -0.259042261 -0.14200871
A15  -0.17355472 -0.259042261 -0.14200871
A19  -0.17355472 -0.259042261 -0.14200871
A21  -0.28415415  0.134917760  0.04258584
A23  -0.28415415  0.134917760  0.04258584
A25  -0.28415415  0.134917760  0.04258584
A28  -0.28415415  0.134917760  0.04258584
A29  -0.28415415  0.134917760  0.04258584
A34  -0.01714403 -0.172125406  0.32197553
A36  -0.01714403 -0.172125406  0.32197553
A37  -0.01714403 -0.172125406  0.32197553
A39  -0.01714403 -0.172125406  0.32197553
A111  0.16195220 -0.084319200 -0.23198014
A13   0.16195220 -0.084319200 -0.23198014
A14   0.16195220 -0.084319200 -0.23198014
A16   0.16195220 -0.084319200 -0.23198014
A17   0.16195220 -0.084319200 -0.23198014
A18   0.16195220 -0.084319200 -0.23198014
A210  0.05135277  0.309640821 -0.04738559
A22   0.05135277  0.309640821 -0.04738559
A24   0.05135277  0.309640821 -0.04738559
A26   0.05135277  0.309640821 -0.04738559
A27   0.05135277  0.309640821 -0.04738559
A33   0.31836290  0.002597655  0.23200410
A35   0.31836290  0.002597655  0.23200410
A38   0.31836290  0.002597655  0.23200410

plotReducedDim(tse, "RDA")
```
Screenshot 2023-08-28 at 10 40 51

RiboRings:
Apparently the reason is that most points are assigned identical coordinates and therefore they overlap in the plot. Is that normal?

TuomasBorman:
Nice!!


> I'm wondering why plotReducedDim only shows 

RDA returns (rda_object$CCA) returns different site scores. Currently, runRDA returns  (Weighted) orthonormal site scores (u), but I guess weighted sums (wa) is more more comparable with other ordination methods. --> we still have to discuss this, see https://github.com/microbiome/mia/pull/422

To overcome this issue while developing this, you can do

```
rda <- reducedDim(tse)

att <- attributes(rda)

rda <- att$rda$CCA$wa

attributes(rda) <- att

reducedDim(tse) <- rda
```

RiboRings:
Thanks! Indeed the wa values seem more meaningful when it comes to plotting.

RiboRings:
Should we make the colours of dots and ellipses consistent? RIght now it's a bit tricky to see which dots belong to which ellipse.

Screenshot 2023-08-28 at 11 44 42

TuomasBorman:
> Should we make the colours of dots and ellipses consistent? RIght now it's a bit tricky to see which dots belong to which ellipse.
> 
> Screenshot 2023-08-28 at 11 44 42

Yes, ggplot_build function or something can probably help to get those colors (take into account that sometimes there is no colours at all, colour_by is not specified)

antagomir:
You could also look at [microshades](https://karstenslab.github.io/microshades/) for accessible color palettes that were suggested for microbiome studies. Seems to support phyloseq but perhaps it is straigfwd to use this also with TreeSE?

RiboRings:
I haven't managed yet to match dot and ellipse colours, because dot are using the scater scale which is manually defined. Maybe I'll try to use `guides`.

Anyway, I think that plot without filling looks better, so I would suggest setting `ellipse_fill = FALSE` by default.

Screenshot 2023-08-28 at 14 57 45

antagomir:
You could also check inspiration from microViz see issue #89

RiboRings:
The colour thing is good to go now. Do you want me to add some optional arguments for these items?

```
# TODO: Ellipse: fill or just an edge? --> Edge line type?
    # TODO: vector: vector line type, vector arrow size and line bulkiness
    # TODO: vector labels: Adjust labels position, turn off ggrepel, text color, text size, text label?, parse off?
    # --> catch all geomrepel parameters (there will be a warning if we try to feed there parameter that doesbeong to plotReducedDim for example)
```

TuomasBorman:
> The colour thing is good to go now. Do you want me to add some optional arguments for these items?
> 
> ```
> # TODO: Ellipse: fill or just an edge? --> Edge line type?
>     # TODO: vector: vector line type, vector arrow size and line bulkiness
>     # TODO: vector labels: Adjust labels position, turn off ggrepel, text color, text size, text label?, parse off?
>     # --> catch all geomrepel parameters (there will be a warning if we try to feed there parameter that doesbeong to plotReducedDim for example)
> ```

Cool, first line is already done. But for those others you could think which option is useful to have. I think at least those are something that the user might want to adjust but there is probably other things also.

RiboRings:
--> turn off ggrepel
Is there a way to turn off ggrepel itself, or do we have to make a conditional and use geom_text instead of geom_text_repel when `repel_text = FALSE`?

--> catch all ggrepel parameters
Should we limit ourselves to the most useful parameters, or is it better to define args for all ggrepel params?

TuomasBorman:
Not sure, but I think this is only option 

> geom_text instead of geom_text_repel when repel_text = FALSE

The problem is that ggrepel leads to an error if other than ggrepel parameters are fed to it. I think there are not too many parameters to not catch them all (but other thoughts are welcome)

RiboRings:
Ok the ggrepel thing is done. Next I saw this comment:

```
# TODO: Catch all plotReducedDim parameters and feed them to plotReducedDim
    # (plotReducedDim cannot accept unknown parameteers, ... is not possibel to use)
```

What do you mean it is not possible to use? Because if for example I mispell `clor_by`, the function still works but doesn't colour the dots, so you can use it if you know which args `plotReducedDim` takes.

TuomasBorman:
> Ok the ggrepel thing is done. Next I saw this comment:
> 
> ```
> # TODO: Catch all plotReducedDim parameters and feed them to plotReducedDim
>     # (plotReducedDim cannot accept unknown parameteers, ... is not possibel to use)
> ```
> 
> What do you mean it is not possible to use? Because if for example I mispell `clor_by`, the function still works but doesn't colour the dots, so you can use it if you know which args `plotReducedDim` takes.

Ok, that is probably just my mistake --> so just feed parameters with `...`

TuomasBorman:
ahh sorry, pushed to wrong branch...

RiboRings:
The plotRDA function is ready now. I tried to update its vignettes but `pkgdown::build_site()` seems to fail:

```
Reading 'man/mia-datasets.Rd'
Reading 'man/mia-plot-args.Rd'
Reading 'man/miaViz-package.Rd'
Reading 'man/plotAbundance.Rd'
Reading 'man/plotAbundanceDensity.Rd'
Reading 'man/plotColTile.Rd'
Reading 'man/plotDMN.Rd'
Reading 'man/plotGraph.Rd'
Reading 'man/plotPrevalence.Rd'
Reading 'man/plotRDA.Rd'
Reading 'man/plotSeries.Rd'
Reading 'man/plotTree.Rd'
Reading 'man/treeData.Rd'
Writing 'reference/treeData.html'
-- Building articles -----------------------------------------------------------
Reading 'vignettes/miaViz.Rmd'
-- RMarkdown error -------------------------------------------------------------

Quitting from lines 36-39 [unnamed-chunk-2] (miaViz.Rmd)
Warning messages:
1: package ‘BiocStyle’ was built under R version 4.3.1 
2: In grSoftVersion() :
  unable to load shared object '/Library/Frameworks/R.framework/Resources/modules//R_X11.so':
  dlopen(/Library/Frameworks/R.framework/Resources/modules//R_X11.so, 6): Library not loaded: /opt/X11/lib/libSM.6.dylib
  Referenced from: /Library/Frameworks/R.framework/Versions/4.3-x86_64/Resources/modules/R_X11.so
  Reason: image not found
3: In cairoVersion() :
  unable to load shared object '/Library/Frameworks/R.framework/Resources/library/grDevices/libs//cairo.so':
  dlopen(/Library/Frameworks/R.framework/Resources/library/grDevices/libs//cairo.so, 6): Library not loaded: /opt/X11/lib/libXrender.1.dylib
  Referenced from: /Library/Frameworks/R.framework/Versions/4.3-x86_64/Resources/library/grDevices/libs/cairo.so
  Reason: image not found
4: In (function (filename = "Rplot%03d.png", width = 480, height = 480,  :
  failed to load cairo DLL
--------------------------------------------------------------------------------
Warning messages:
1: package ‘MatrixGenerics’ was built under R version 4.3.1 
2: package ‘IRanges’ was built under R version 4.3.1 
3: package ‘GenomeInfoDb’ was built under R version 4.3.1 
4: package ‘MultiAssayExperiment’ was built under R version 4.3.1 
Error: 
! in callr subprocess.
Caused by error in `map(.x, .f, ..., .progress = .progress)`:
! In index: 1.
ℹ See `$stdout` and `$stderr` for standard output and error.
Type .Last.error to see the more details.
```

RiboRings:
It might be just because the Cairo package does not work on Mac, does it build for you?

R/plotCCA.R
```diff
@@ -0,0 +1,359 @@
...
+    )
+    # Get scatter plot with plotReducedDim --> keep theme similar between ordination methods
+    plot <- do.call("plotReducedDim", plotReducedDim_args)
+    
+    # Get data for ellipse
+    ellipse_data <- NULL
+    if( add_ellipse && !is.null(colour_by) ){
+        ellipse_data <- reduced_dim
+        ellipse_data <- as.data.frame(ellipse_data)
+        ellipse_data[[colour_by]] <- colData(tse)[[colour_by]]
```
TuomasBorman, line 101: Replace with scater::retriveCellInfo

TuomasBorman:
That might be the problem, vignettes build just fine in my computer


The method looks robust --> now we need plotCCA alias and documents need to be updated. In case of plotRecucedDim and geom_repel paramaters, you can just add tem to additional paramaters and say that `...` is fed to these functions

RiboRings:
I updated the docs, but I am not sure what they look like because I can't compile them in my laptop.

I noticed that the method works only for se/tse objects and not for rda/cca objects, but maybe we can add the rda/cca method at some later point?

TuomasBorman:
I can check when time allows

> I noticed that the method works only for se/tse objects and not for rda/cca objects, but maybe we can add the rda/cca method at some later point?

Actually, my plan was to also implement this so that it works with calculateRDA and rda object. Because you also ended up thinking this, we could implement it

My suggestion:

Because we want that these plots looks similar despite what is the input class

1. Create TreeSE from rda object. (Get sample names --> create an empty assay --> add reducedDim)
2. Plot this TreeSE --> with calculateRDA input you could keep significance parameters as they are but with rda object input, disable significance parameter since they are not included

--> Make this work with as little code as possible so that the "real workhorse" is the code that is already implemented

RiboRings:
Everything ready for your final check.

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+setMethod("plotCCA", signature = c(object = "SingleCellExperiment"),
+    function(object, dimred, ...){
+        ###################### Input check #######################
+            
+        ###################### Input check end ####################
+        # Get data for plotting
+        plot_data <- .incorporate_rda_vis(object, dimred, ...)
+        # Create a plot
+        plot <- .rda_plotter(plot_data, ...)
+        return(plot)
```
TuomasBorman, line 123: This could be just return(plotRDA(object, ...))

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+#' @aliases plotRDA
+#' @export
+setGeneric("plotCCA", signature = c("object"),
+           function(object, dimred, ...) standardGeneric("plotCCA"))
+
+#' @rdname plotRDA
+#' @aliases plotRDA
+#' @export
+setMethod("plotCCA", signature = c(object = "SingleCellExperiment"),
+    function(object, dimred, ...){
```
TuomasBorman, line 115: This could be just `function(object, ...){`

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+    }
+)
+
+
+################## HELP FUNCTIONS ##########################
+
+# Construct TreeSE from rda/cca object to pass it to downstream functions
+.rda2tse <- function(object) {
+    # If rda/cca object, convert it to TreeSE
+    if (is.matrix(object)) {
```
TuomasBorman, line 191: The class is already checked in the generic function (runRDA)

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+        # Remove significance attribute if present
+        attr(rda, "significance") <- NULL
```
TuomasBorman, line 193: How about including this? So just keep it and put the matrix as it is to reducedDim? If the input is matrix, it should be from calculateRDA (if it is not, then .incorporate_rda_vis should handle the issue

TuomasBorman: Also there is no rda variable

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+        # Make empty colData
+        coldata <- data.frame(ID = rownames(rda))
```
TuomasBorman, line 195: How about DataFrame(row.names = c("asd", "lfjf"))

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+        rownames(coldata) <- coldata[["ID"]]
+        coldata[["ID"]] <- NULL
+        # Convert rda/cca object to TreeSE
```
TuomasBorman, line 198: Since assay is only required one, you could make that to simplify the code

Something like this matrix(ncol = 3, dimnames = list(NULL, c("sam1", "sam2", "sam3")))

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+#'   \code{\link[scater:plotReducedDim]{plotReducedDim}},
+#'   \code{\link[ggplot2:geom_label]{geom_label}} and
+#'   \code{\link[ggrepel:geom_label_repel]{geom_label_repel}}.
+#' 
+#' @details
+#' plotRDA and plotCCA create an RDA/CCA plot starting from the output of
+#' \code{\link[mia:runCCA]{CCA and RDA}} functions, two common methods for
+#' supervised ordination of microbiome data. Both
+#' \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
+#' and rda/cca objects are supported as input. When the input is a SummarizedExperiment
```
TuomasBorman, line 56: The input must be SingleCellExperiment (SE does not have reducedDim slot)

Also rda object is not supported. Matrix is supported (which is an output of calculateRDA)

TuomasBorman: rda object can be found from the matrix attributes though...

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
+#' Plot RDA or CCA object
+#'
+#'
+#' @param object a
+#'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}, a
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-constructor]{TreeSummarizedExperiment}}
+#'   or a rda/cca object. The latter is the output of \code{\link[mia:runCCA]{calculateRDA}}.
```
TuomasBorman, line 7: SingleCellExperiment. You could just state here TreeSUmmarizedExperiment for consistency (and matrix)

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+#' 
+#' @param dimred A string or integer scalar indicating the reduced dimension to
+#'   plot. This is the output of \code{\link[mia:runCCA]{runRDA}} and resides in
+#'   \code{reducedDim(tse, dimred)}.
+#'
+#' @param colour_by String specifying the name of a column in colData to colour by
+#'   (Default: NULL).
+#' 
+#' @param color_by Alias for `colour_by`.
```
TuomasBorman, line 16: These are parameters of plotReducedDim I guess? Then I think it is better to remove them from here

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+#' 
+#' @param ellipse_fill if TRUE, ellipses are filled. (Default: TRUE).
+#'
+#' @param alpha Number between 0 and 1 to adjust the opacity of ellipses (Default: 0.2).
```
TuomasBorman, line 20: opacity of ellipse fill

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+#' @param dimred A string or integer scalar indicating the reduced dimension to
+#'   plot. This is the output of \code{\link[mia:runCCA]{runRDA}} and resides in
+#'   \code{reducedDim(tse, dimred)}.
+#'
+#' @param colour_by String specifying the name of a column in colData to colour by
+#'   (Default: NULL).
+#' 
+#' @param color_by Alias for `colour_by`.
+#' 
+#' @param ellipse_fill if TRUE, ellipses are filled. (Default: TRUE).
```
TuomasBorman, line 18: For consistency, check other function documents. Boolean values are described somehow like "TRUE or FALSE specifying whether to fill ellipses" or something. Try to keep these all descriptions similar

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+#' @aliases plotCCA
+#' @export
+setGeneric("plotRDA", signature = c("object"),
+    function(object, dimred, ...) standardGeneric("plotRDA"))
+
+#' @rdname plotRDA
+#' @aliases plotCCA
+#' @export
+setMethod("plotRDA", signature = c(object = "SingleCellExperiment"),
+    function(object, dimred, ...){
```
TuomasBorman, line 155: All plotRDA specific parameters (listed above) should be listed here as well

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+        ###################### Input check #######################
+
```
TuomasBorman, line 157: And here input checks of those parameters

R/plotCCA.R
```diff
@@ -0,0 +1,504 @@
...
+#'  
+#'  # Create RDA plot without repelling text
+#'  plotRDA(tse, "RDA",
+#'          colour_by = "ClinicalStatus,
+#'          repel_text = FALSE)
+#'  
+#'  
+NULL
+
+#' @rdname plotRDA
```
TuomasBorman, line 105: rdname should be plotCCA --> the name of document

R/plotCCA.R
RiboRings: This throws an error if changed. Isn't `dimred` also a compulsory arg?

TuomasBorman: I was thinking if plotCCA does not have any parameters than `object` --> so there is as little as possible and everything happens in plotRDA

TuomasBorman: We can consider moving dimred parameter to `.incorporate_rda_vis`

However, because it is always needed, it might be easier to be in plotRDA since then user can just type plotRDA(tse, "RDA") and not plotRDA(tse, dimred = "RDA")

RiboRings: Yes however if `dimred` is not included also in the args of `plotCCA`, it throws an error:

```

plotCCA(tse, "RDA")
# Error in plotRDA(object, ...) :
# argument "dimred" is missing, with no default
```

antagomir: Whatever the solution will be we should make sure that it is systematic across the family of similar functions.

I do not see great immediate need for supporting dimred here but there can be use cases (e.g. using dimension reduction as a rough denoising for the data).

TuomasBorman: --> yes, we can keep it here in this case

RiboRings:
I think it can be merged now

DESCRIPTION
```diff
@@ -46,7 +49,9 @@ Imports:
     tidyr,
     dplyr,
     ape,
-    DirichletMultinomial
+    DirichletMultinomial,
+    ggrepel,
+    SingleCellExperiment
```
TuomasBorman, line 54: Is this SingleCellExperiment dependency needed? Is mia dependency enough? --> less dependencies are better

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+#' @param vec_color Alias for `vec_colour`.
+#' 
+#' @param vec_linetype Discrete number specifying the style of vector lines.
+#'   (default: \code{vec_linetype = 1})
+#' 
+#' @param arrow_size Number specifying the size of arrows.
+#'   (defaults: \code{arrow_size = 0.25})
+#' 
+#' @param label_size Number specifying the size of text and labels.
+#'   (default: \code{label_size = 4})
```
TuomasBorman, line 42: Sorry... one thing came to my mind. We are switching from `para_meter `to `para.meter` in miaverse. So these underscores should be changed to dots, I just remembered....

functionName
parameter.name

See issue: https://github.com/microbiome/mia/issues/405

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+#' 
+#' @details
+#' \code{plotRDA} and \code{plotCCA} create an RDA/CCA plot starting from the
+#' output of \code{\link[mia:runCCA]{CCA and RDA}} functions, two common methods
+#' for supervised ordination of microbiome data. Both
+#' \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-constructor]{TreeSummarizedExperiment}}
+#' and matrix objects are supported as input. When the input is a
+#' TreeSummarizedExperiment, this should contain the output of runRDA
+#' in the reducedDim slot and the argument \code{dimred} needs to be defined.
+#' When the input is an matrix, this should be returned as output from
```
TuomasBorman, line 68: a matrix

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+#' calculateRDA and the argument \code{dimred} needs not be defined. However,
```
TuomasBorman, line 69: You could remove "and the argument \code{dimred} needs not be defined."  because it is not in matrix function definition --> I think more clear by removing from here

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+#' # Load dataset
+#' library(miaViz)
+#' data("enterotype", package = "mia")
+#' tse <- enterotype
+#' 
+#' # Remove missing values from covariables
+#' tse <- tse[ , !is.na(tse$ClinicalStatus)]
+#' tse <- tse[ , !is.na(tse$Age)]
+#' tse <- tse[ , !is.na(tse$Gender)]
+#'
```
TuomasBorman, line 89: I think na.action parameter is more clear to use --> we can modify runRDA function and catch the `na.action` parameter. Without the parameter error occurs, which might be quite hard to solve. --> we could make a check and give informative messages (and maybe modify the default behavior of na.action)

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-constructor]{TreeSummarizedExperiment}}
+#'   or a matrix of weights. The latter is returned as output from \code{\link[mia:runCCA]{calculateRDA}}.
+#' 
+#' @param dimred A string or integer scalar indicating the reduced dimension to
+#'   plot. This is the output of \code{\link[mia:runCCA]{runRDA}} and resides in
+#'   \code{reducedDim(tse, dimred)}.
+#' 
+#' @param ellipse_fill TRUE or FALSE, should ellipses be opacity filled.
+#'   (default: \code{ellipse_fill = TRUE})
+#'
```
TuomasBorman, line 17: There is also add_ellipse parameter (which is hidden now) --> You could combine these two parameters --> maybe add_ellipse which can be TRUE, FALSE, "fill" and "color"/"colour". (TRUE could be same as "color". With FALSE, no ellipses are plotted

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+
+################## HELP FUNCTIONS ##########################
+
+# Construct TreeSE from rda/cca object to pass it to downstream functions
+#' @importFrom S4Vectors DataFrame
+.rda2tse <- function(object) {
+    # Convert rda/cca object to TreeSE
+    object <- TreeSummarizedExperiment(
+      assays = matrix(ncol = nrow(object), dimnames = list(NULL, rownames(object))),
+      colData = DataFrame(row.names = rownames(object)),
```
TuomasBorman, line 237: colData is not needed to define, so you can delete this line

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+        if( any(ind) ){
+            # Get biplot from cca object
+            rda <- attributes(reduced_dim)[ind][[1]]
+            vector_data <- rda$CCA$biplot
+            vector_data <- as.data.frame(vector_data)
+            vector_data[["group"]] <- rownames(vector_data)
+        } else{
+            # If it cannot be found, give warning
+            warning(
+                "CCA object was not found. Please calculate CCA by using runCCA.",
```
TuomasBorman, line 301: The function also supports RDA and calculateRDA. Maybe "CCA/RDA object was not found. Please calculate it by using run*/calculate* function."

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+            signif_data <- as.data.frame(signif_data)
+            # Add significance to vector labels
+            # Get vector labels
+            vector_label <- vector_data[["vector_label"]]
+            # Add significance to vector labels
+            vector_label <- .add_signif_to_vector_labels(vector_label, variable_names, signif_data, ...)
+            vector_data[["vector_label"]] <- vector_label
+        } else{
+            # If it cannot be found, give warning
+            warning("Significance data was not found. please calculate CCA by using runCCA.",
```
TuomasBorman, line 360: Same here: also calculate* is possible

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+        xlab = xlab,
+        ylab = ylab
+    )
+    return(result)
+}
+
+# Make vector labels more tidy, i.e, separate variable and group names.
+# Replace also underscores with space
+.tidy_vector_labels <- function(
+        vector_label, coldata, sep_group = "\U2012", sep_underscore = " ", ...){
```
TuomasBorman, line 401: This is just internal stuff, but you could still add input checks for `sep_group `and `sep_underscore`. Better name for sep_underscore would be repl_underscore since underscore is replaced

R/plotCCA.R
```diff
@@ -0,0 +1,540 @@
...
+        }
+        # Replace underscores with space
+        new_name <- gsub("_", sep_underscore, new_name)
+        return(new_name)
+    })
+    return(vector_label)
+}
+
+# This function adds significance info to vector labels
+.add_signif_to_vector_labels <- function(vector_label, var_names, signif_data, sep_underscore = " ", ...){
```
TuomasBorman, line 426: Same here, sep_underscore

TuomasBorman (commented):
Good, minor things anymore

Also: bump the version from DESCRIPTION and add info to NEWS file

TuomasBorman:
I think this is very nice

Do you have something to add @antagomir

antagomir:
Very good. Once this is merged then make sure OMA examples are also updated.

DESCRIPTION
RiboRings: I added it as a dep because the CMD check was complaining about `#' @importFrom SingleCellExperiment reducedDim reducedDimNames`, since `reducedDim` and `reducedDimNames` are used in the `.incorporate_rda_vis` helper function.

Is there a better way to do this?

TuomasBorman: Ahh ok, good --> then this is ok, I guess there is no better way to do this

This does not really add dependencies since SCE is already dependency of mia

R/plotCCA.R
RiboRings: Do you mean that we could control both colour and fill with only one argument (add_ellipse)?

But then ellipses without fill (which look nice) cannot be produced.

RiboRings: Ah never mind, now I got it!

NEWS
```diff
@@ -19,4 +19,7 @@ Changes in version 1.7.x
 + Fixed plotGraph* that was broked due changes in dependencies
 
 Changes in version 1.9.x
-+ Updated plotDMN to work with newest mia version
\ No newline at end of file
++ Updated plotDMN to work with newest mia version
+
+Changes in version 1.10.x
```
TuomasBorman, line 24: This is still version 1.9.x

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: miaViz
 Title: Microbiome Analysis Plotting and Visualization
-Version: 1.9.2
```
TuomasBorman, line 3: Correct version is 1.9.3

TuomasBorman:
Forgot to say, you could still add tests to https://github.com/microbiome/miaViz/tree/master/tests/testthat

RiboRings:
I thought the examples in the vignettes were working as tests. Or should I add tests to https://github.com/microbiome/miaViz/tree/master/tests/testthat anyway?

TuomasBorman:
Yes, every time we create or modify a function, we should create also tests that test that method is working as expected. That is probably the most boring part but ensures that everything is working (and also if we do some modifications later [or if there are changes in dependencies], we get an alert if the functionality of method is changed)

Tests should include
- Check that error occurs if wrong argument value
- Take the ggplot object. Check that values are correct. (`ggplot_build` function might help)
- Check that ellipses, vectors etc are not plotted if they are FALSE

RiboRings:
Hi! I finished writing some tests.

Not really sure what this error in GH Actions is all about:

```
Run Rscript -e "remotes::install_github('r-hub/sysreqs')"
   File /usr/local/lib/R/etc//Renviron.site contains invalid line(s)
      <html>
      <head><title>301 Moved Permanently</title></head>
      <body>
      <center><h1>301 Moved Permanently</h1></center>
      <hr><center>CloudFront</center>
      </body>
      </html>
   They were ignored
Using github PAT from envvar GITHUB_PAT
Skipping install of 'sysreqs' from a github remote, the SHA1 (f068afa9) has not changed since last install.
  Use `force = TRUE` to force installation
/bin/bash: eval: line 12: syntax error near unexpected token `('
/bin/bash: eval: line 12: `   File /usr/local/lib/R/etc//Renviron.site contains invalid line(s)      <html>      <head><title>301 Moved Permanently</title></head>      <body>      <center><h1>301 Moved Permanently</h1></center>      <hr><center>CloudFront</center>      </body>      </html>   They were ignoredexport DEBIAN_FRONTEND=noninteractive; apt-get -y update && apt-get install -y pandoc make libxml2-dev libicu-dev libgmp-dev libglpk-dev'
Error: Process completed with exit code 2.
```
It looks external or?

TuomasBorman:
Same is in mia. I don't know what is the problem but it is not related to our packages -->  it is related to the environment where these tests are run

TuomasBorman:
I might have to merge this to master since I need to use this function today in one demo and it probably would be too complicated to say to load the package from branch... I open an issue about this

TuomasBorman:
btw superb work!!!!

TuomasBorman:
I ran the checks locally, test-plotCCA.R:48:3 seems to give an error

TuomasBorman:
> I ran the checks locally, test-plotCCA.R:48:3 seems to give an error

fixed

RiboRings:
Thanks no problem

PR #105, opened 2023-10-19
replace deprecations with new funcs

ake123:


R/plotDMN.R
```diff
@@ -55,7 +55,7 @@ setMethod("plotDMNFit", signature = c(x = "SummarizedExperiment"),
             dmn <- metadata(x)[[name]]$dmm
         } else {
             .Deprecated(old="getDMN", new="cluster", 
-                    "Now runDMN and calculateDMN are deprecated. Use cluster with DMMParam parameter and full parameter set as true instead.")
+                    "Now runDMN and cluster are deprecated. Use cluster with DMMParam parameter and full parameter set as true instead.")
```
TuomasBorman, line 58: Here is error

TuomasBorman: Also add msg="Now runDMN...." --> the deprecation message is not printed correctly (I have noticed that in different functions)

antagomir:
@ake123 could you finalize this one so we can close?

tests/testthat/test-2plotGraph.R
```diff
@@ -50,7 +50,7 @@ test_that("plot graph", {
     expect_equal(miaViz:::.colnames_tbl_graph(actual,"edges"),
                  c("from","to","weight"))
     #
-    genus <- agglomerateByRank(GlobalPatterns,"Genus",na.rm=TRUE)
```
TuomasBorman, line 53: This is incorrect. mergeFeaturesByRank will be removed

TuomasBorman:
This PR replaces agglomerateByRank with mergeFeaturesByRank, however, the mergeFeaturesByRank will be deprectated. (I have not updated Bioc devel of mia since we have still some modifications in GH that are under discussion)

Also, the cluster function will be deprecated so this PR has only old information. Sorry for that

Issue #86, opened 2023-08-12
argument names


antagomir:
Argument `order_rank_by` could be better replace with `order_feature_by` following other similar changes across the mia framework

0xMuluh:
Addressed in #134 .

PR #120, opened 2024-04-25
changed order_rank_by to order_features_by

ndegunst:
Updated the parameter name, see issue #86.


R/plotAbundance.R
```diff
@@ -30,7 +30,7 @@
 #'   character will be plotted as colour-code bar. (default: \code{features =
 #'   NULL})
 #'   
-#' @param order_rank_by How to order abundance value: By name (\dQuote{name}) 
+#' @param order_features_by How to order abundance value: By name (\dQuote{name})
```
antagomir, line 33: should this be order_rows_by, or is this coming later in parameter renaming scheme (@TuomasBorman )?

antagomir (approved):

antagomir:
@Insaynoah what's the status of this one?

ndegunst:
I completely forgot about this one. It should be working. The checks failed because it still tries to run the tests that are currently in use and don't have the updated function name. I updated the test files within the PR and seem to run just fine.

R/plotAbundance.R
TuomasBorman: Sorry, I did not see this PR. We are planning to rename this. It is not following the scheme that we are going to use. --> parameter.name

antagomir: Should we change here now or is that gonna be a separate PR (potentially from @Daenarys8 )?

TuomasBorman: I think it is better to change these all at once. We have already discussed about this with @Daenarys8

antagomir: Does this mean that we can merge and close this PR then?

TuomasBorman: I had now time to check this in detail. There are couple of problems:

1. This is not deprecating the old name which means that the might be unexpected behavior if user does not notice this.
2. The parameter name should be order.rows.by


So these things should be either addressed before merging, or then other option is to close this PR and make these changes in other PR

antagomir:
Hmm still some warnings & errors...

R/plotAbundance.R
ndegunst: I'll close this one and make a new PR, that seems easier.

Issue #110, opened 2024-04-02
plotRDA ellipse's confidence level


TuomasBorman:
Documentation does not say about the confidence level of ellipse. (plotRDA uses stat_ellipse with default level which is 0.95) Add information to documentation. Consider adding parameter to specify level.

antagomir:
This should be propagated to some OMA example, too?

PR #121, opened 2024-05-07
Adding confidence level as a parameter

ElySeraidarian:
Issue #110


R/plotCCA.R
```diff
@@ -24,6 +24,9 @@
 #' 
 #' @param ellipse.linetype Discrete number specifying the style of ellipses.
 #'   (default: \code{ellipse.linetype = 1})
+#'   
+#' @param ellipse.level Number between 0 and 1 to adjust confidence level.
+#'   (default: \code{ellipse.level = 0.95})
```
antagomir: wouldn't a more descriptive name be "confidence.level"?

antagomir (commented):
Looks good to me overall. One comment on argument naming.

antagomir:
status of this?

ElySeraidarian:
Fixed what had to be, however build still fails in ubuntu and no idea how to solve (gave it a few try already)

TuomasBorman:
This should be work when psuhed to devel

TuomasBorman:
Also bump the version

TuomasBorman:
Also update NEWS file "Added confidence.interval parameter to plotCCA" or something

TuomasBorman (approved):

ElySeraidarian:
Should be fine now

R/plotCCA.R
```diff
@@ -225,6 +228,9 @@ setMethod("plotRDA", signature = c(object = "SingleCellExperiment"),
         if ( !is.numeric(ellipse.linewidth) && ellipse.linewidth > 0 ) {
             stop("'ellipse.linewidth' must be a positive number.", call. = FALSE)
         }
+        if( confidence.level < 0 || confidence.level > 1 ){
+          stop("'confidence.level' must be a number between 0 and 1.", call. = FALSE)
+        }
```
TuomasBorman: Ahh sorry, I did not send the comment. Can you add is.numeric()? That will ensure that TRUE and FALSE will be not passed through

R/plotCCA.R
```diff
@@ -228,7 +228,7 @@ setMethod("plotRDA", signature = c(object = "SingleCellExperiment"),
         if ( !is.numeric(ellipse.linewidth) && ellipse.linewidth > 0 ) {
             stop("'ellipse.linewidth' must be a positive number.", call. = FALSE)
         }
-        if( confidence.level < 0 || confidence.level > 1 ){
+        if( is.numeric(confidence.level) < 0 || is.numeric(confidence.level) > 1 ){
```
TuomasBorman: This is not correct since is.numeric() outputs logical.

--> !(is.numeric(confidence.interval) && confidence.interval > 0 && confidence.interval < 1)

ElySeraidarian: Yes sorry

TuomasBorman: confidence.level = TRUE still passes through without error. Also I noticed that other input check are also incorrect. There are multiple ways to think about this if-sentence but I think the most simplest is to think first the supported values and then just take negation -->

!(is.numeric(confidence.interval) && confidence.interval > 0 && confidence.interval < 1)

... which means that the value must be numeric and between 0 and 1. If it is not, then it give TRUE.

--> Can you fix these input checks from this function so that they follow this?

TuomasBorman (approved):

PR #125, opened 2024-05-29
changed order_rank_by to order.rows.by with deprecation

ndegunst:
Here's the new PR following PR #120. This now includes deprecation warnings and the new argument name is order.rows.by.


TuomasBorman:
This is little bit too complex solution. Moreover, we are planning to rename these parameter all at once, but I think you could do that.

We have similar thing going on in mia currently: https://github.com/microbiome/mia/pull/562

I can give you more info in the following days since I have not decided the names of parameters for miaViz. Then we can have a short chat about the renaming. Does this sound good plan?

Issue #96, opened 2023-09-15
Strange example


antagomir:
The first example on the [plotAbundance](https://microbiome.github.io/miaViz/reference/plotAbundance.html) reference page asks to plot "counts" assay but the figure is relative abundances. This makes the example confusing.

Explain the default behavior, if the relabundance plotting is automated as a default; or (even better?) update the figure so that it plots by default what is asked (here, counts)

SHillman836:
https://github.com/microbiome/miaViz/pull/126 - for the change

antagomir:
@SHillman836 ok so this issue can be closed?

SHillman836:
yes

antagomir:
Great. I will close this now. Do actively close issues that are done.

PR #126, opened 2024-06-04
Strange example #96 - Changing miaViz documentation example

SHillman836:


antagomir:
Did you edit the file in man/ directory directly? 

The documentation uses roxygen2 and these texts should be changed in the corresponding R/ file headers. The man/ contents are automatically generated with devtools::document()

antagomir:
instead of `ylab()` use `labs(y=...)` (that's nowaday the recommended one)

SHillman836:
> Did you edit the file in man/ directory directly?
> 
> The documentation uses roxygen2 and these texts should be changed in the corresponding R/ file headers. The man/ contents are automatically generated with devtools::document()

Ah ok - yes I did (sorry never worked with this before), I'll make that change now and edit the y axis calling

PR #127, opened 2024-06-04
updated example in docs

SHillman836:


antagomir (commented):
Looks good. Have you tested the other checks mentioned in How to develop the package section in [contributor guidelines](https://github.com/microbiome/mia/blob/devel/CONTRIBUTING.md)?

The devtools::build(), devtools::check(), devtools::install() and then the Bioconductor checks? Those must pass without warnings or errors.

TuomasBorman (approved):

.Rbuildignore
```diff
@@ -1,3 +1,5 @@
+^renv$
+^renv\.lock$
```
TuomasBorman, line 2: Can you describe briefly why these were changed?

SHillman836: Hey, I just added these mainly because I thought the renv didn't need to be a part of the package build - just as you wouldn't want it being pushed up to github - is that not correct? I can change it back if needed

TuomasBorman: That is good. Just wanted to be sure what was the purpose.

SHillman836: Ok nice sounds good.

PR #128, opened 2024-06-13
Adding sorting method

SHillman836:


R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+#' 
+#' @examples
+#' # Load required libraries
+#' library(mia)
+#' 
+#' # Load the dataset
+#' data(peerj13075)
+#' 
+#' # Agglomerate by Order and transform the data
+#' tse_order <- mergeFeaturesByRank(peerj13075, rank = "order", onRankOnly = TRUE)
```
TuomasBorman: agglomerateByRank is the new name of the function. Update your mia

R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+#' @return A vector of row names in the sorted order.
+#' @rdname neatsort
+#' @export
+setMethod("neatsort", signature = c("matrix"),
+          function(x,
+                   subset = NULL,
+                   dimensions = c(1, 2),
+                   centering_method = c("mean", "median", "none"),
+                   sorting_order = c("ascending", "descending"),
+                   ...){
```
TuomasBorman: The indentation should be multiplication of 4 spaces

TuomasBorman: We use this naming convention:

- ClassName
- functionName
- .internal_function_name
- parameter.name

R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+                return(ordering)
+          }
+    )
+
+
+# Takes a subset of rows from the data matrix.
+.take_subset <- function(data, subset) {
+    data <- data[subset, , drop = FALSE]
+    return(data)
+}
```
TuomasBorman: Avoid this kind of functions. As this is just one line, this function is overkill

R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+
+
+# Takes the specified columns (dimensions) from the data matrix.
+.take_dimensions <- function(data, dimensions) {
+    data <- data[, dimensions, drop = FALSE]
+    return(data)
+}
```
TuomasBorman: Same here, just add the line directly to "main" code

R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+    
+    theta <- atan2(centered_data[, 2], centered_data[, 1])
+    names(theta) <- rownames(centered_data)
+    
+    return(theta)
+}
+
+# Sorts the theta values and returns the ordered row names.
+.get_sorted_rownames <- function(theta_values, sorting_order) {
+    sorted_indices <- order(theta_values, decreasing = (sorting_order == "descending"))
```
TuomasBorman: You could rename the parameter to decreasing which get values TRUE or FALSE. That would make the code simpler. Also as the same parameter is used in base R, it would make this more intuitive

R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+    # Check for unique row names
+    if (any(duplicated(rownames(x)))) {
+        stop("Row names of the matrix must be unique.")
+    }
+    
+    # Check for unique column names
+    if (any(duplicated(colnames(x)))) {
+        stop("Column names of the matrix must be unique.")
+    }
+}
```
TuomasBorman: You could add this internal function below the generic methods ( as they are in most of the files)

R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+        } else if (is.character(subset) && any(!subset %in% rownames(x))) {
+            stop("Subset refers to row names that do not exist in the data.")
+        } else if (!is.numeric(subset) && !is.character(subset)) {
+            stop("Subset must be a vector of row indices or names.")
+        }
+    }
+    
+    # Check dimensions are valid
+    if (any(dimensions > ncol(x))) {
+        stop("dimensions refer to columns that do not exist in the data.")
```
TuomasBorman: In stop() and warnings() add `call. = FALSE`. That makes the message tidier

R/neatsort.R
```diff
@@ -0,0 +1,186 @@
...
+    
+    # Check for unique row names
+    if (any(duplicated(rownames(x)))) {
+        stop("Row names of the matrix must be unique.")
+    }
+    
+    # Check for unique column names
+    if (any(duplicated(colnames(x)))) {
+        stop("Column names of the matrix must be unique.")
+    }
```
TuomasBorman: As R functions always return something, it is good practice to specify the returned value. Here it could be just return(NULL)

TuomasBorman (changes requested):
Looks good, but should be in mia

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' 
+#' One significant application of this sorting method is in plotting heatmaps. 
+#' By using radial theta sorting, the relationships between data points can be preserved 
+#' according to the ordination method's spatial configuration, rather than relying on 
+#' hierarchical clustering, which may distort these relationships. This approach 
+#' allows for a more faithful representation of the data's intrinsic structure as captured 
+#' by the ordination process.
+#' 
+#' @param x A matrix containing the ordinated data to be sorted. Columns should represent the principal components (PCs) and rows should represent the entities being analyzed (e.g., features or samples).
+#' @param subset A vector specifying a subset of rows to be used and retained. If NULL, all rows are used.
```
antagomir: Do we need this parameter - what is the added value compared to just inserting subsetted data?

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' @param dimensions A vector of length 2 specifying the columns of the matrix to use for the X and Y coordinates.
```
antagomir: same question; it would simplify this function if we expect user to do something like `getNeatOrder(x[subset, dimensions], ...)` instead of  `getNeatOrder(x, subset, dimensions, ...)`. In the former case this function does not have to deal with the subsetting part.

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' @param centering_method A character string specifying the method to center the data. Options are "mean", "median", or "none" if your data is already centred.
```
antagomir: use `centering.method`

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' @param decreasing A boolean that when true sorts the rows in a descending order by radial theta angle. Default is False.
```
antagomir: should be `FALSE` (not `False`)

antagomir: write "(default: FALSE)"

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' @examples
+#' # Load required libraries
+#' library(mia)
+#' 
+#' # Load the dataset
+#' data(peerj13075)
+#' 
+#' # Agglomerate by Order and transform the data
+#' tse_order <- agglomerateByRank(peerj13075, rank = "order", onRankOnly = TRUE)
+#' tse_order <- transformAssay(tse_order, assay.type = "counts", method="relabundance", MARGIN = "samples", name="relabundance")
```
antagomir: you could add explanation comment: "Transforms the samples into relative abundances"

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' tse_order <- transformAssay(tse_order, assay.type="relabundance", method="z", MARGIN = "features", name="z")
```
antagomir: you could add explanation comment: "Transforms the features (taxa) into zero mean, unit variance (z transformation)"

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' z_transformed_data <- assay(tse_order, "z")
+#' 
+#' # Get the top taxa and perform PCA
+#' top_taxa <- getTopFeatures(tse_order, top = 10, assay.type="z")
```
antagomir: I think this is better to remove. Usually PCA would be done with all features in this context unless there are scalability issues. Agglomerating the data to the order level should already take care of that (dimensionality should be relatively low for fast calculation).

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' top_feature_data <- z_transformed_data[top_taxa, ]
+#' pca_results <- prcomp(top_feature_data, scale = TRUE)
+#' scores_pca <- pca_results$x[, 1:2]
```
antagomir: We should use methods from mia / miaViz to perform the ordination in the mia examples. See getPCA or addPCA

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' 
+#' # Sort by radial theta and subset the transformed data
+#' sorted_order <- getNeatOrder(scores_pca, dimensions = c(1, 2), centering_method = "mean")
+#' ordered_transformed_data <- z_transformed_data[sorted_order, ]
```
antagomir: I would sort the TreeSE object directly with `tse[, sorted_order]` and then use the heatmap tools that support SummarizedExperiment (sechm?).

antagomir:
Looks good overall but I added some comments to sync with the rest of the package

R/getNeatOrder.R
TuomasBorman: I agree. For easier maintenance for the future (and also to make the function easier to use) the function should only things that are necessary

R/getNeatOrder.R
TuomasBorman: How about just `method`?

R/getNeatOrder.R
TuomasBorman: \code{FALSE} so that it looks nicer in documentaton. Check other documenation pages

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+#' centers the data using specified methods (mean, median, or none), and then calculates 
+#' the angle (theta) for each point relative to the centroid. The data points are then 
+#' sorted based on these theta values in either ascending or descending order. 
+#' 
+#' One significant application of this sorting method is in plotting heatmaps. 
+#' By using radial theta sorting, the relationships between data points can be preserved 
+#' according to the ordination method's spatial configuration, rather than relying on 
+#' hierarchical clustering, which may distort these relationships. This approach 
+#' allows for a more faithful representation of the data's intrinsic structure as captured 
+#' by the ordination process.
```
TuomasBorman: Very clear

antagomir:
See also: https://github.com/microbiome/OMA/issues/146

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+    
+    # Check for unique row names
+    if (any(duplicated(rownames(x)))) {
+        stop("Row names of the matrix must be unique.", call. = FALSE)
+    }
+    
+    # Check for unique column names
+    if (any(duplicated(colnames(x)))) {
+        stop("Column names of the matrix must be unique.", call. = FALSE)
+    }
```
TuomasBorman: These input check could be added directly to "main" code since there are not too many parameters to check

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+.radial_theta <- function(data, centering_method) {
+    if (centering_method == "mean") {
+        centered_data <- scale(data, center = TRUE, scale = FALSE)
+    } else if (centering_method == "median") {
+        centered_data <- scale(data, center = apply(data, 2, median), scale = FALSE)
+    } else if (centering_method == "none") {
+        centered_data <- data
+    } else {
+        stop("Unsupported centering method. Choose either 'mean', 'median', 'mode', or 'none'.", call. = FALSE)
+    }
```
TuomasBorman: This could be simplified. Something like this probably

center_fun <- switch(method, "median" = median, "mean" = mean)
if( !is.null(center_fun) ){
    center_vals <- apply(data, 2, center_fun)
    data <- data - center_vals
}

TuomasBorman: Also add comments

R/getNeatOrder.R
```diff
@@ -0,0 +1,179 @@
...
+}
+
+# Sorts the theta values and returns the ordered row names.
+.get_sorted_rownames <- function(theta_values, decreasing) {
+    sorted_indices <- order(theta_values, decreasing = decreasing)
+    rownames <- names(theta_values)[sorted_indices]
+    return(rownames)
+}
+
+
```
TuomasBorman: One extra line

tests/testthat/test-getNeatOrder.R
```diff
@@ -0,0 +1,203 @@
...
+    
+    # Test with valid inputs and subset row names
+    subset <- c("Sample1", "Sample3")
+    result <- miaViz::getNeatOrder(neatsort_matrix, subset, dimensions = c(1, 2), centering_method = "mean")
+    expected <- c("Sample3", "Sample1")
+    expect_equal(result, expected)
+    
+    # Test with valid inputs and subset row indices
+    subset <- c(1, 3)
+    result <- miaViz::getNeatOrder(neatsort_matrix, subset, dimensions = c(1, 2), centering_method = "mean")
```
TuomasBorman: Call functions directly without specifying the package

miaViz::getNeatOrder --> getNeatOrder

R/getNeatOrder.R
TuomasBorman: The neatsort paper could be cited

TuomasBorman (changes requested):
Looks good!

At this point, I would like to discuss about the TreeSE/SCE support.

For example [mia::calculateJSD](https://github.com/microbiome/mia/blob/devel/R/calculateJSD.R) have support for both matrix and SE.

Similarly, this could support TreeSE. So in addition to this current implementation, there could be method for TreeSE which has parameters for selecting

- dimension reduction (dimred parameter)
- number of dimensions (ncomponents)
- whether to sort the output TreeSE (sort)

However, reducedDim slot requires that number of columns/samples match. If I understood correctly, this "neatsort" can be done also for features which means that reducedDim slot cannot be utilized (scater::calculatePCA and scater::calculateMDS can still be used as they have transposed parameter and they return only the output matrix).  If that is the case, then I think that we should only support method for matrix to make things simpler.

antagomir:
> For example [mia::calculateJSD](https://github.com/microbiome/mia/blob/devel/R/calculateJSD.R) have support 
> for both matrix and SE. Similarly, this could support TreeSE. 

Support for SE should be enough for methods that do not use the tree information. Because TreeSE is SE (it inherits SE). 

Currently we do not seem to have ordination methods that utilize the tree. Those exist I think but they are not commonly used. It is an interesting possibility but probably not very urgent.

> So in addition to this current implementation, there could be method for TreeSE which has parameters for selecting 
> * dimension reduction (dimred parameter)
> * number of dimensions (ncomponents)
> * whether to sort the output TreeSE (sort)

Adding dimension reduction would mean that we combine the steps or ordination and sorting? Possible but at leas it would complicate the implementation because now the sorting can be done based on any ordination method output. If these are combined, we need to choose which ordination methods are supported. If I got this right.

I am not sure if the radial theta angle methods is defined beyond the 2D case, hence ncomponents might not be valid option w.r.t. neatsort.

Sorting can be also done afterwards, the added value might be limited.
 
> However, reducedDim slot requires that number of columns/samples match. If I understood correctly, this "neatsort" can be done also for features which means that reducedDim slot cannot be utilized (scater::calculatePCA and scater::calculateMDS can still be used as they have transposed parameter and they return only the output matrix). If that is the case, then I think that we should only support method for matrix to make things simpler.


Neatsort can be done for any Nx2 numeric matrix. Usually it is applied to sample scores is samples x PCaxes score matrix but in principle other uses are possible. 

TBH it might not be critical to support feature loading matrices (features x PCaxes) because this doesn't seem very common. But it is possible and potentially useful, and out current implementations seems to have that support because it can operate with arbitrary Nx2 matrices.

Indeed, we could choose to simply support sample scores and use the reducedDim slot for that. This will complicate the implementation and the added value is not yet clear. I tend to think that we could keep getNearOrder support for matrices for now. If real use cases will call for a more streamlined implementation and TreeSE support we can always add that. But then we will need to address these more complex design questions.

This is somewhat related to the question on how we should present feature loadings in general in TreeSE objects because reducedDim is indeed only supporting the sample scores, and missing the feature loadings. The metadata slot can be used for that at least.

TuomasBorman:
I meant that user could first do dimension reduction with methods that add the result to reducedDim and then call getNeatOrder(). As SE does not have reducedDim slot, at least SCE should have been supported.

But I agree, that only matrix should be supported at this point. This would make the usage simple

```
res <- calculatePCA(tse)
order <- getNeatOrder(res)
tse <- tse[ , order]
sechm(tse)
```

antagomir:
Right, I agree: getNearOrder could support reducedDim (TreeSE/SCE). But perhaps best to think about that later since the cost/benefit ratio seems a bit open.

SHillman836:
Thanks both for the review and comments, I'm working on all of these updates today!

SHillman836:
So I have implemented the changes but I have a few follow up questions just before I push.

1) In my example I've used the runPCA method to create the PCA matrix. This is because it was the only method I could see in OMA - (https://microbiome.github.io/OMA/docs/devel/pages/20_beta_diversity.html#sec-other-ord-methods). Is that OK? 

2) Then also when calling sechm in my example am I right that we do the PCA, get the ordering from getNeatOrder, apply that to the tse object... but when we plot with sechm - do we plot using the z-transformed data? 

3) When citing the publication, does that go in @details? Or @references? Or just slipped into the main description?

4) Please also see my responses to this OMA issue - https://github.com/microbiome/OMA/issues/146

5) As per Tuomas' suggestion I've simplified my centering logic.  However, Tuomas suggested doing the subtractions manually, but originally I'd use the scale method. Which should I do? As I understand it the scale method can be used to do centering more precisely, and which one I choose does make a slight difference.

TuomasBorman:
1. More appropriate might be calculatePCA which returns matrix directly. We could highlight better where functions come from so that user can refer to those package manuals

2. Sounds correct. Z-transformation is usually applied to heatmaps since the colors are sensitive to scales. If there are features that varies a lot compared to others in "absolute" scale, those "others" look like that they do not vary at all. When values are scaled by features, the colors are scaled by feature (and not by whole dataset). If I remember correctly sechm has scale parameter, but it might be better to do manually (plot with Z-transformed data)

3.  References https://github.com/microbiome/mia/blob/devel/R/calculateUnifrac.R

4. Noted. Leo probably knows better

5. I did not know that. Use scale(), but simplify the structure. There was at least one extra if-else

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
 #' allows for a more faithful representation of the data's intrinsic structure as captured 
 #' by the ordination process.
 #' 
-#' @param x A matrix containing the ordinated data to be sorted. Columns should represent the principal components (PCs) and rows should represent the entities being analyzed (e.g., features or samples).
-#' @param subset A vector specifying a subset of rows to be used and retained. If NULL, all rows are used.
-#' @param dimensions A vector of length 2 specifying the columns of the matrix to use for the X and Y coordinates.
-#' @param centering_method A character string specifying the method to center the data. Options are "mean", "median", or "none" if your data is already centred.
-#' @param decreasing A boolean that when true sorts the rows in a descending order by radial theta angle. Default is False.
+#' @param x A matrix containing the ordinated data to be sorted. Columns should represent the principal components (PCs) and rows should represent the entities being analyzed (e.g. features or samples).
+#' @param dimensions A \code{character} or \code{integer} vector of length 2 specifying the columns of the matrix to use for the X and Y coordinates.
```
antagomir: do we need this argument?

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
+#' @param centering.method A single \code{character} value specifying the method to center the data. Options are \code{mean}, \code{median}, or \code{none} if your data is already centered. (default: method = \code{mean})
```
antagomir: how about just "method"

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
+#' @param decreasing A \code{boolean} that when \code{TRUE} sorts the rows in a descending order by radial theta angle. (default: descending = \code{FALSE})
 #' @param ... Additional arguments passed to other methods.
-#' @return A vector of row names in the sorted order.
+#' @return A \code{character} vector of row names in the sorted order.
+#' 
+#' @details 
+#' It's important to note that the sechm package does actually have the functionality for plotting a heatmap using this radial theta angle ordering, though only by using an MDS ordination.
```
antagomir: write \pkg{sechm}

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
+#' 
+#' This functionality can be found at:
+#' 
+#'  \url{https://bioconductor.org/packages/3.18/bioc/vignettes/sechm/inst/doc/sechm.html#row-ordering}. 
+#'  
+#' That being said, the \code{getNeatOrder} function is more modular and separate to the plotting, and 
+#' can be applied to any kind of ordinated data which can be valuable depending on the use case.
+#' 
+#' @references
+#' The below paper outlines the NeatMap method in more detail:
```
antagomir: Usual way to do is to include line 38 in the @details section and cite it with "Rajaram & Oono (2010)"

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
+#' 
+#' Rajaram, S., Oono, Y. NeatMap - non-clustering heat map alternatives in R. BMC Bioinformatics 11, 45 (2010). https://doi.org/10.1186/1471-2105-11-45.
+#' 
+#' It can be found at:
+#' 
+#'  \url{https://bmcbioinformatics.biomedcentral.com/articles/10.1186/1471-2105-11-45}.
```
antagomir: remove l 42-44, the doi is already providing the permanent link

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
-#' scores_pca <- pca_results$x[, 1:2]
-#' 
-#' # Sort by radial theta and subset the transformed data
-#' sorted_order <- getNeatOrder(scores_pca, dimensions = c(1, 2), centering_method = "mean")
-#' ordered_transformed_data <- z_transformed_data[sorted_order, ]
+#' ## Group data by taxonomic order
+#' tse <- agglomerateByRank(peerj13075, rank = "order", onRankOnly = TRUE)
+#' 
+#' ## Transform the samples into relative abundances
+#' tse <- transformAssay(tse, assay.type = "counts", method="relabundance", MARGIN = "samples", name="relabundance")
```
antagomir: better to write MARGIN="cols"

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
+#' tse <- transformAssay(tse, assay.type="relabundance", method="z", MARGIN = "features", name="z")
+#' 
+#' ## Perform PCA using calculatePCA
+#' pca_result <- calculatePCA(tse, ncomponents = 10, assay.type = "z")
+#' 
+#' ## Add PCA results to the TreeSE object
+#' reducedDim(tse, "PCA") <- pca_result
+#' 
+#' ## Sort by radial theta and sort the original assay data
+#' sorted_order <- getNeatOrder(reducedDim(tse, "PCA"), dimensions = c(1, 2), centering.method = "mean")
```
antagomir: this should be just 
sorted_order <- getNeatOrder(reducedDim(tse, "PCA")[, c(1,2)], method = "mean")

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
+#' tse <- tse[, sorted_order]
+#' 
+#' ## Create the heatmap with sechm whilst retaining this radial theta ordering
+#' features <- rownames(assay(tse, "z"))
```
antagomir: isnt this the same than `rownames(tse)` ?

-> simplify (if possible)?

R/getNeatOrder.R
```diff
@@ -17,38 +17,64 @@
...
+#' 
+#' ## Perform PCA using calculatePCA
+#' pca_result <- calculatePCA(tse, ncomponents = 10, assay.type = "z")
+#' 
+#' ## Add PCA results to the TreeSE object
+#' reducedDim(tse, "PCA") <- pca_result
+#' 
+#' ## Sort by radial theta and sort the original assay data
+#' sorted_order <- getNeatOrder(reducedDim(tse, "PCA"), dimensions = c(1, 2), centering.method = "mean")
+#' tse <- tse[, sorted_order]
```
antagomir, line 74: The use of this with heatmap is not demonstrated, whereas sechm is. 

It would be better to show heatmap with getNeatOrder+heatmap, or both that one and sechm (for comparison). But in function examples you are supposed to primarily  demonstrate how that function (getNeatOrder) is used to support heatmap plotting.

R/getNeatOrder.R
SHillman836: Ok will do

R/getNeatOrder.R
SHillman836: I can remove decreasing if needed yeah, it would mean the sorting would only occur in one order but I can definitely do that

R/getNeatOrder.R
antagomir: Note that the name "z" was just changed to "scale" in the new mia version:
https://github.com/microbiome/mia/issues/583

-> Update this here as well?

R/getNeatOrder.R
SHillman836: @antagomir - just to confirm - you want me to remove decreasing, and should I remove the dimensions argument too?

SHillman836: cause if I was to call like this - sorted_order <- getNeatOrder(reducedDim(tse, "PCA")[, c(1,2)], method = "mean") - that would mean removing the dimensions argument. Although I quite like it as an argument cause you can make it default (1,2) - but it's very easy to also just do it when you call the method

R/getNeatOrder.R
antagomir: Hi - sorry to confuse - perhaps just "centering"? ("method" can be less clear and in general i prefer shorter argument names as long as they are clear enough)

R/getNeatOrder.R
antagomir: The question is more like, this argument is not needed as 1) it is faster to write x[, c(1,2)] in input than "x, dimensions=c(1,2)" in the function call and 2) this will also simplify function and reduce maintenance burden (many small things like this will create a big task stream in the end)

R/getNeatOrder.R
SHillman836: @antagomir which mia version is this? I'm on 1.12.0 and i can't seem to update it from that version, and scale isn't recognized on this version. Neither is margin="cols"

antagomir: very recent mia PR that was merged
https://github.com/microbiome/mia/pull/584

-> in fact I didn't check if they updated the version number.

OK the margin you can keep as is. I think that wasn't merged yet.

SHillman836: Ah ok yeah, I can add in scale, although it's not quite working yet in RStudio as my mia can't update to the recent PR. Also, is this suitable in terms of the heatmap my example creates?

Screenshot 2024-06-20 at 12 32 48

SHillman836: The samples have the ordering from getNeatOrder applied to them. Will wrap this up and the OMA chapter today btw so I can move onto the EBI stuff

antagomir: Ok this looks better. You could check the following however:

1) confirm manually that the rows (taxa) have zero mean, unit variance (they should, after "scale")
2) you can consider capping the colors at for instance to [-5, 5] (values exceeding these would be capped to -5 and 5, respectively; this is a common visualization practice and ok when transparently reported
3) I realize that the starting sample can be decided arbitrarily, there is no way to determine the "correct" first sample; you could test how changing the start sample (left most sample) would impact the figure. Here, the left and right ends would form a continuum as well (as the angle rotates full circle)

SHillman836: ok thanks will do. yes all the data has near 0 mean and stdev of 1

R/getNeatOrder.R
```diff
@@ -0,0 +1,173 @@
...
+#' assay(tse, "counts") <- assay(tse, "counts") + 1
+#' 
+#' ## Transform the samples into relative abundances using CLR
+#' tse <- transformAssay(tse, assay.type = "counts", method="clr", MARGIN = "samples", name="clr")
+#' 
+#' ## Transform the features (taxa) into zero mean, unit variance (z transformation)
+#' tse <- transformAssay(tse, assay.type="clr", method="z", MARGIN = "features", name="z")
+#' 
+#' ## Perform PCA using calculatePCA
+#' reducedDim(tse, "PCA") <- calculatePCA(tse, ncomponents = 10, assay.type = "z")
```
TuomasBorman: You can skip this part. Just get the result to new variable. The idea was to make the example simpler.

TuomasBorman: Also ncomponents is unnecessary here, I guess

TuomasBorman (changes requested):

R/getNeatOrder.R
TuomasBorman: res <- calculatePCA(tse, assay.type = "z")
order <- getNeatOrder(res, ...)

SHillman836: OK will do, yeah I can understand that this could be seen as simpler, but then again I also think it's just personal preference at a certain point. Cause I code in Rust a fair bit you have to manage memory quite a lot, so I generally avoid creating excess variable in my code, as this is less memory efficient, so for me I find saving stuff to an object simpler than creating excess variables.

SHillman836: Yeah sure I can change ncomponents, although then again it's just like if you were doing this in reality, the default is 500 components, so you would probably cap the ncomponents at a certain point as you just dont want to deal with that many components

TuomasBorman: To store the result to TreeSE, you can use runPCA. However, as these are examples of usage of getNeatOrder(), we should make the code that is not directly related to this function as simple as possible

For instance, we have found that reducedDim might be hard to understand first, so it is better to make the examples simple so it is easy for all users to see what is happening.

TuomasBorman: One point of using TreeSE is the memory management. However, OMA is more suitable for more detailed information and examples  (including best practices)

SHillman836: ok sure will do, I get you

SHillman836: Was that the only change by the way, making those changes quickly now, lmk if there's anything else to change, I can't really see anything else with the code but there may be small things you notice I don't

R/getNeatOrder.R
```diff
@@ -0,0 +1,173 @@
...
+.radial_theta <- function(data, centering) {
+    # Choose the correct centering function based on the method
+    center_fun <- switch(centering, "median" = median, "mean" = mean)
+    
+    # Apply the centering if there's a method present
+    if (!is.null(center_fun)) {
+        center_vals <- apply(data, 2, center_fun)
+        centered_data <- scale(data, center = center_vals, scale = FALSE)
+    } else if (centering == "none") {
+        centered_data <- data
```
TuomasBorman: Instead of centered_data variable, you could just modify data variable

tests/testthat/test-getNeatOrder.R
```diff
@@ -0,0 +1,137 @@
+testthat::set_max_fails(Inf)
+
```
TuomasBorman: Remove extra line (1 is enough)

tests/testthat/test-getNeatOrder.R
```diff
@@ -0,0 +1,137 @@
+testthat::set_max_fails(Inf)
```
TuomasBorman: I think it is not necessary to define this?

R/getNeatOrder.R
```diff
@@ -0,0 +1,173 @@
...
+#' 
+#' One significant application of this sorting method is in plotting heatmaps. 
+#' By using radial theta sorting, the relationships between data points can be preserved 
+#' according to the ordination method's spatial configuration, rather than relying on 
+#' hierarchical clustering, which may distort these relationships. This approach 
+#' allows for a more faithful representation of the data's intrinsic structure as captured 
+#' by the ordination process.
+#' 
+#' @param x A matrix containing the ordinated data to be sorted. Columns should represent the principal components (PCs) and rows should represent the entities being analyzed (e.g. features or samples). There should be 2 columns only representing 2 PCs.
+#' @param centering A single \code{character} value specifying the method to center the data. Options are \code{mean}, \code{median}, or \code{none} if your data is already centered. (default: method = \code{mean})
```
TuomasBorman: --> (default: \code{"mean"})

R/getNeatOrder.R
```diff
@@ -0,0 +1,173 @@
...
+#' @param ... Additional arguments passed to other methods.
+#' @return A \code{character} vector of row names in the sorted order.
+#' 
+#' @details 
+#' It's important to note that the \pkg{sechm} package does actually have the functionality for plotting a heatmap using this radial theta angle ordering, though only by using an MDS ordination. 
+#' 
+#' This functionality can be found at:
+#' 
+#'  \url{https://bioconductor.org/packages/3.18/bioc/vignettes/sechm/inst/doc/sechm.html#row-ordering}.
```
TuomasBorman: You could add this with [link](link address)

R/getNeatOrder.R
```diff
@@ -0,0 +1,173 @@
...
+#' This functionality can be found at:
+#' 
+#'  \url{https://bioconductor.org/packages/3.18/bioc/vignettes/sechm/inst/doc/sechm.html#row-ordering}. 
+#'  
+#' That being said, the \code{getNeatOrder} function is more modular and separate to the plotting, and 
+#' can be applied to any kind of ordinated data which can be valuable depending on the use case.
+#' 
+#' The below paper outlines the NeatMap method in more detail:
+#' 
+#' "Rajaram & Oono (2010)" - https://doi.org/10.1186/1471-2105-11-45.
```
TuomasBorman: Also this could be added with (Rajaram & Oono (2010) NeatMap - non-clustering heat map alternatives in R)[https://doi.org/10.1186/1471-2105-11-45]

--> Check the manual age from man folder

R/getNeatOrder.R
```diff
@@ -0,0 +1,173 @@
...
+#' data(peerj13075)
+#' 
+#' ## Group data by taxonomic order
+#' tse <- agglomerateByRank(peerj13075, rank = "order", onRankOnly = TRUE)
+#' 
+#' ## Add a pseudocount to the counts data
+#' assay(tse, "counts") <- assay(tse, "counts") + 1
+#' 
+#' ## Transform the samples into relative abundances using CLR
+#' tse <- transformAssay(tse, assay.type = "counts", method="clr", MARGIN = "samples", name="clr")
```
TuomasBorman: You can add pseudocount with pseudocount = TRUE in transformAssay

TuomasBorman:
Run BiocCheck::BiocCheck(). There were some NOTEs related to these lines. Resolve those notes that can reasonably do. (If line contains only URL address, you can keep that even though it exceeds 80 characters)

R/getNeatOrder.R
```diff
@@ -0,0 +1,164 @@
...
+#' [Rajaram & Oono (2010) NeatMap - non-clustering heat map alternatives in R](https://doi.org/10.1186/1471-2105-11-45) outlines this in more detail.
+#' 
+#' @name getNeatOrder
+#' 
+#' @examples
+#' ## Load the required libraries and dataset
+#' library(mia)
+#' library(scater)
+#' library(ComplexHeatmap)
+#' library(circlize)
```
TuomasBorman, line 53: Where is this package needed

R/getNeatOrder.R
```diff
@@ -0,0 +1,164 @@
...
+#'               name = "NeatMap",
+#'               col = col_fun,
+#'               cluster_rows = FALSE,  # Do not cluster rows
+#'               cluster_columns = FALSE,  # Do not cluster columns
+#'               show_row_dend = FALSE,
+#'               show_column_dend = FALSE,
+#'               row_names_gp = gpar(fontsize = 4), 
+#'               column_names_gp = gpar(fontsize = 6), 
+#'               heatmap_width = unit(20, "cm"),  
+#'               heatmap_height = unit(15, "cm")
```
TuomasBorman: Is there a reason to not use sechm package to simplify the plotting?

R/getNeatOrder.R
```diff
@@ -0,0 +1,164 @@
...
+#' 
+#' [Rajaram & Oono (2010) NeatMap - non-clustering heat map alternatives in R](https://doi.org/10.1186/1471-2105-11-45) outlines this in more detail.
+#' 
+#' @name getNeatOrder
+#' 
+#' @examples
+#' ## Load the required libraries and dataset
+#' library(mia)
+#' library(scater)
+#' library(ComplexHeatmap)
```
TuomasBorman, line 52: Packages used in the examples must be added to Suggests in DESCRIPTION file.

TuomasBorman (changes requested):
Looks good! Adding ComplexHeatmap to Suggests probably fixes the error

R/getNeatOrder.R
SHillman836: circlize is needed for the colorRamp2 function

R/getNeatOrder.R
SHillman836: ok thanks will do

R/getNeatOrder.R
SHillman836: I think Leo had asked in a previous comment to use ComplexHeatmap, with this in mind should I still change the OMA example plot to sechm?

R/getNeatOrder.R
TuomasBorman: Thanks!

R/getNeatOrder.R
TuomasBorman: ComplexHeatmap is fine here. I think in OMA we should use sechm whenever it is possible to streamline the analysis (for example plotting abundances). However, sechm is not suitable for every situation which is why we have to have some examples on ComplexHeatmap also (for example when plotting correlation heatmap). The nice thing is that sechm is just a wrapper for ComplexHeatmap so both approaches give the result in same format

SHillman836: ok thanks sounds good

TuomasBorman:
Resolve errors in unit test

SHillman836:
What're the errors - there was definitely no errors when I last checked

antagomir:
"All checks have failed" (see above Github runs)

antagomir:
> test_check("miaViz")
[ FAIL 1 | WARN 6 | SKIP 1 | PASS 175 ]

══ Skipped tests (1) ═══════════════════════════════════════════════════════════
• miaTime cannot be loaded (1): 'test-2plotSeries.R:5:5'

══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-getNeatOrder.R:48:5'): Test check_args method ────────────────
`.check_args(check_args_matrix[, c(1, 2)], "invalid_method")` threw an error with unexpected message.
Expected match: "'arg' should be one of “mean”, “median”, “none”"
Actual message: "'arg' should be one of \"mean\", \"median\", \"none\""
Backtrace:
    ▆
 1. ├─testthat::expect_error(...) at test-getNeatOrder.R:48:5
 2. │ └─testthat:::quasi_capture(...)
 3. │   ├─testthat (local) .capture(...)
 4. │   │ └─base::withCallingHandlers(...)
 5. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
 6. └─miaViz:::.check_args(check_args_matrix[, c(1, 2)], "invalid_method")
 7.   └─base::match.arg(centering, c("mean", "median", "none"))

[ FAIL 1 | WARN 6 | SKIP 1 | PASS 175 ]
Error: Test failures
Execution halted

3 errors ✖ | 0 warnings ✔ | 4 notes ✖
Error: Process completed with exit code 1.
Run actions/upload-artifact@v3
/usr/bin/docker exec  01b1f8ffe51deb581bc7323b620ee9945bab51cd97bc196a49f233c188b27d66 sh -c "cat /etc/*release | grep ^ID"

SHillman836:
ok interesting, I missed that cause it passed locally with devtools::test(), but failed in the CI/CD pipeline. Anyway I've made the testing abit more robust so will commit it now, which should sort it out.

R/getNeatOrder.R
```diff
@@ -0,0 +1,165 @@
...
+    # Check centering argument
+    centering <- match.arg(centering, c("mean", "median", "none"))
+    # Check for unique row names
+    if (any(duplicated(rownames(x)))) {
+        stop("Row names of the matrix must be unique.", call. = FALSE)
+    }
+    # Check for unique column names
+    if (any(duplicated(colnames(x)))) {
+        stop("Column names of the matrix must be unique.", call. = FALSE)
+    }
```
TuomasBorman: Is unique column names required? I think they are not in this implementation.

Also can we work with index instead of rownames. That would make the function more robust. Can you check?

TuomasBorman: That is what order() outputs

```
> order(c("A", 1, 4, 3, 5, 3))
[1] 2 4 6 3 5 1
```

TuomasBorman: Otherwise, the function is good to go

SHillman836: ok yeah thanks, just about to push and fix it will be a few mins

TuomasBorman: Can you check if rownames are necessary?

SHillman836: As in do you mean if we need the rowname uniqueness check? It works with both rownames and indices, and returns sorted row indices not rownames now.

TuomasBorman: Yes, in theory TreeSE can have duplicated rownames. Can you chrck that the function works if there are no rownames or if they are duplicated

TuomasBorman: Although matrix might require rownames, cannot remember by heart

TuomasBorman: But it is good to make the function as general as possible. Usually it is better to work with row and column indices instead of names. Index is always unique.

SHillman836: updated it now, yeah it works w no rownames and with duplicates

TuomasBorman (approved):

TuomasBorman:
Great work! Looks nice

TuomasBorman:
Ubuntu check:

Error in `eval(code, test_env)`: object 'expected_indices' not found


Mac and windows fails because they do not find new mia functions. They should be resolved automatically

TuomasBorman:
Fixed the error. I also changed the allowed values for centering to be mean, median and NULL (NULL is common to denote that something has been disabled)

SHillman836:
OK great, thanks, sounds good!

PR #130, opened 2024-06-17
Orgnanise parameter names

0xMuluh:
This commit updates parameter naming to follow the `parameter.name` convention


R/plotAbundance.R
```diff
@@ -24,36 +24,45 @@
 #'   (Please use \code{assay.type} instead. At some point \code{assay_name}
 #'   will be disabled.)
 #'   
-#' @param features a single \code{character} value defining a column from 
+#' @param cor.var a single \code{character} value defining a column from
```
TuomasBorman: features is not good name since usually we mean taxa/rows with features.

However, this cor.var sounds like "variable that is analyzed for correlation" or something. col.var might be better --> "column variable".

R/plotAbundance.R
```diff
@@ -476,10 +490,10 @@ MELT_VALUES <- "Value"
...
                             xlab,
                             flipped,
-                            add_legend,
-                            add_x_text,
-                            point_alpha,
-                            point_size){
+                            add.legend,
+                            add.x.text,
+                            point.alpha,
+                            point.sizze){
```
TuomasBorman: point.sizze

R/plotAbundanceDensity.R
```diff
@@ -20,26 +20,34 @@
...
 #'   results be ordered in a descending order? If \code{NA} is given the order
-#'   as found in \code{object} for the \code{n} most abundant taxa is used.
-#'   (default: \code{order_descending = TRUE})
+#'   as found in \code{x} for the \code{n} most abundant taxa is used.
+#'   (default: \code{order.descending = TRUE})
+#'   
+#' @param order_descending Deprecated. Use \code{order.descending} instead.
 #'   
 #' @param ... additional parameters for plotting. 
 #' \itemize{
```
TuomasBorman, line 53: You should be careful with additional parameters. In some functions, we use pass them to scater package. Should be checked that all these modifying parameters are in miaViz, and these modifications works (the unit tests should check that though),  Just wanted to comment so that you are aware of that

R/plotColTile.R
```diff
@@ -145,15 +145,15 @@ setMethod("plotRowTile", signature = c("SummarizedExperiment"),
 }
 
 .tile_plotter <- function(data,
-                          add_legend = TRUE,
+                          add.legend = TRUE,
                           xlab,
```
TuomasBorman: Check that the old parameter still works (if that can be modified by user)

R/plotGraph.R
```diff
@@ -199,30 +213,36 @@ setGeneric("plotRowGraph", signature = c("x","y"),
...
+             edge.colour.by = edge_colour_by,
              edge_colour_by = NULL,
+             edge.width.by = edge_width_by,
              edge_width_by = NULL,
              colour_by = NULL,
+             shape.by = shape_by,
              shape_by = NULL,
+             size.by = size_by,
              size_by = NULL,
-             by_exprs_values = "counts",
```
TuomasBorman, line 211: It seems that this parameter was completely removed and not deprecated?

R/plotGraph.R
```diff
@@ -541,42 +567,42 @@ setMethod("plotRowGraph",
...
-                           size_by = NULL,
+                           shape.by = NULL,
+                           size.by = NULL,
                            line_alpha = 1,
                            line_width = NULL,
                            line_width_range = c(0.5,3),
-                           point_alpha = 1,
-                           point_size = 2,
+                           point.alpha = 1,
+                           point.sizze = 2,
```
TuomasBorman: Does the old parameter names still work?

TuomasBorman: point_sizze

R/plotPrevalence.R
```diff
@@ -28,45 +28,57 @@
 #'   (Please use \code{assay.type} instead. At some point \code{assay_name}
 #'   will be disabled.)
 #'   
-#' @param as_relative logical scalar: Should the detection threshold be applied
+#' @param as.relative logical scalar: Should the detection threshold be applied
 #'   on compositional (relative) abundances? Passed onto
 #'   \code{\link[mia:getPrevalence]{getPrevalence}}. (default: \code{TRUE})
 #'   
-#' @param colour_by Specification of a feature to colour points by, see the 
+#' @param colur.by Specification of a feature to colour points by, see the
```
TuomasBorman: colur.by

TuomasBorman (changes requested):
In general looks good. However, many parameters are just renamed without supporting the old parameter name. We have to ensure that support for old parameters still stays.

0xMuluh:
okay. i will add general support for old param names.

0xMuluh:
Hmm, in an attempt to support old param names in internal function, a lot of bugs have been created. Will prepare a v2 for this PR which is cleaner.

TuomasBorman:
Is there any progress on this "version 2"?

0xMuluh:
Still a WIP, contemplating on ignoring all `.functions`  param and only update `funcTions` (fxns in docs).

PR #133, opened 2024-07-03
Update MARGIN param

0xMuluh:
This commit performs the following change(reference: [#587](https://github.com/microbiome/mia/pull/587));

- samples --> cols
- features --> rows


TuomasBorman (approved):

TuomasBorman:
mia must first be pushed to Bioc devel; I will do it, will take couple of days to update

PR #134, opened 2024-07-08
Update params in user facing fxns

0xMuluh:
This commit updates parameters of user functions.


R/plotAbundance.R
```diff
@@ -148,12 +157,17 @@ setGeneric("plotAbundance", signature = c("x"),
...
             rank = taxonomyRanks(x)[1],
+            col.var = features,
             features = NULL,
+            order.row.by = order_rank_by,
             order_rank_by = c("name","abund","revabund"),
+            order.col.by = order_sample_by,
             order_sample_by = NULL,
             decreasing = TRUE,
+            use.relative = use_relative,
             use_relative = TRUE,
```
TuomasBorman: use.relative might be moved to hidden parameter --> consider

R/plotAbundanceDensity.R
```diff
@@ -20,26 +20,34 @@
...
 #'   \code{colData}, that is used to group observations to different point size
-#'   groups. Must be a value of \code{colData()} function. \code{size_by} is
-#'   disabled when \code{layout = "density"}. (default: \code{size_by = NULL})
+#'   groups. Must be a value of \code{colData()} function. \code{size.by} is
+#'   disabled when \code{layout = "density"}. (default: \code{size.by = NULL})
+#'   
+#' @param size_by Deprecated. Use \code{size.by} instead.
 #' 
-#' @param order_descending \code{TRUE}, \code{FALSE} or\code{NA}: Should the
+#' @param order.descending \code{TRUE}, \code{FALSE} or\code{NA}: Should the
```
TuomasBorman: This could be just decreasing = TRUE as in base::order()

R/plotPrevalence.R
```diff
@@ -149,26 +161,26 @@ setMethod("plotPrevalence", signature = c(x = "SummarizedExperiment"),
...
+        if(as.relative && (any(detections < 0) || any(detections > 1))){
+            stop("If 'as.relative' == TRUE, detections' must be numeric ",
                  "values between 0 and 1.", call. = FALSE)
         }
         #
         x <- mia:::.agg_for_prevalence(x, rank, ...)
         plot_data <- .get_prevalence_plot_data(x, assay.type, detections,
-                                               prevalences, as_relative,
+                                               prevalences, as.relative,
                                                BPPARAM)
```
TuomasBorman: Same here, as.relative could be moved to hidden parameters

TuomasBorman (changes requested):
Initial thoughts

R/plotAbundance.R
0xMuluh: I agree, since `use_relative` is passed directly to `.get_abundance_data`

R/plotAbundanceDensity.R
0xMuluh: ACK.

R/plotAbundance.R
TuomasBorman: I made a larger PR that improved abundance and prevalence plotting functions. In the PR, I already took into account the issue related to moving use_relative to hidden parameter.

https://github.com/microbiome/miaViz/pull/113

TuomasBorman: Prevalence plotting functions are using as.relative. Abundance potting is using use.relative. We must harmonize this --> let's use as.relative also here

R/plotAbundance.R
```diff
@@ -24,47 +24,55 @@
...
 #'   used to define the behavior of the scales of each facet. Both values are 
 #'   passed onto \code{\link[ggplot2:facet_wrap]{facet_wrap}}.
 #' 
 #' @param ... additional parameters for plotting.
 #'   \itemize{
-#'   \item{use_relative}{ \code{TRUE} or \code{FALSE}: Should the relative values
-#'   be calculated? (default: \code{use_relative = FALSE} }
+#'   \item{as.relative}{ \code{TRUE} or \code{FALSE}: Should the relative values
+#'   be calculated? (default: \code{as.relative = FALSE} }
+#'   \item{use_relative} Deprecated. Use \code{as.relative} instead.
```
TuomasBorman: You can remove use_relative from this documentation. The lack of this internal parameters does not cause checks to fail. Also, if user is using the old code, it is still working even though we are not promoting it here

TuomasBorman:
plotPrevalence has parameter called label. plotGraph has the same parameter called show.label. I think they are doing the same thing, but "label" is little bit more unclear. -> can you check that they do the same thing and rename them to show.label?

R/plotTree.R
```diff
@@ -5,25 +5,33 @@
...
 #'   alphabetic order of taxonomic levels?  
-#'   (default: \code{order_tree = FALSE})
+#'   (default: \code{order.tree = FALSE})
+#'   
+#' @param order_tree Deprecated. Use \code{order.tree} instead.
+#'   
+#' @param remove.levels logical scalar, Should taxonomic level information
+#'   be removed from labels? (default: \code{relabel.tree = FALSE})
 #'   
-#' @param remove_levels logical scalar, Should taxonomic level information
```
TuomasBorman, line 23: This could be levels.rm

tests/testthat/test-2plotTree.R
TuomasBorman, line 1: It seems that you have updated only these tests. I guess the other functions should be also updated?

R/plotPrevalence.R
```diff
@@ -67,17 +77,25 @@
...
+#'   are calculated automatically. \code{as.relative} is then also regarded as 
 #'   \code{TRUE}.
 #' 
 #' @param BPPARAM A
 #'   \code{\link[BiocParallel:BiocParallelParam-class]{BiocParallelParam}}
 #'   object specifying whether the UniFrac calculation should be parallelized.
+#'   
+#' @param ... additional parameters for plotting. 
+#' \itemize{
+#'  \item{as_relative} \code{Boolean} indicating whether the detection threshold
```
TuomasBorman: Should be as.relative

tests/testthat/test-2plotTree.R
0xMuluh: ok noted.

R/plotTree.R
```diff
@@ -167,128 +202,168 @@
...
-                              show_highlights = show_highlights,
-                              show_highlight_label = show_highlight_label,
-                              abbr_label = abbr_label,
-                              add_legend = add_legend,
+setMethod("plotColTree", signature = c(x = "TreeSummarizedExperiment"),
+          function(x,
+                   tree.name = tree_name,
+                   tree_name = "phylo",
+                   relabel.tree = relabel_tree,
+                   relabel_tree = FALSE,
```
TuomasBorman: Can you check indentation here since these lines are already modified?

0xMuluh:
> plotPrevalence has parameter called label. plotGraph has the same parameter called show.label. I think they are doing the same thing, but "label" is little bit more unclear. -> can you check that they do the same thing and rename them to show.label?

Hmm, `label` focuses on rownames while `show.label` handles logical vectors for selective labeling and matches character vectors to node data columns i.e `label` is used to specify which labels from the rownames of the data should be selected and `show.label` controls whether and which tip labels are plotted.

Should I go ahead and rename to `show.lable`?

TuomasBorman:
> > plotPrevalence has parameter called label. plotGraph has the same parameter called show.label. I think they are doing the same thing, but "label" is little bit more unclear. -> can you check that they do the same thing and rename them to show.label?
> 
> Hmm, `label` focuses on rownames while `show.label` handles logical vectors for selective labeling and matches character vectors to node data columns i.e `label` is used to specify which labels from the rownames of the data should be selected and `show.label` controls whether and which tip labels are plotted.
> 
> Should I go ahead and rename to `show.lable`?

I think the idea is still the same; select labels that are going to be shown in the plot? Right? I think then the names should be equal (at least for me it was not intuitive what `label` do. For `show.label` it is much more clear what it does; controls labels)

0xMuluh:
> > > plotPrevalence has parameter called label. plotGraph has the same parameter called show.label. I think they are doing the same thing, but "label" is little bit more unclear. -> can you check that they do the same thing and rename them to show.label?
> > 
> > 
> > Hmm, `label` focuses on rownames while `show.label` handles logical vectors for selective labeling and matches character vectors to node data columns i.e `label` is used to specify which labels from the rownames of the data should be selected and `show.label` controls whether and which tip labels are plotted.
> > Should I go ahead and rename to `show.lable`?
> 
> I think the idea is still the same; select labels that are going to be shown in the plot? Right? I think then the names should be equal (at least for me it was not intuitive what `label` do. For `show.label` it is much more clear what it does; controls labels)

I agree. `show.label` is more intuitive. Will add the change.

R/plotAbundanceDensity.R
```diff
@@ -121,88 +129,92 @@
...
-#' # for custom ordering set order_descending = NA and order the input object
+#' # for custom ordering set decreasing = NA and order the input object
 #' # to your wishes
 #' plotAbundanceDensity(
-#'     tse, assay.type = "relabundance", order_descending = NA)
+#'     tse, assay.type = "relabundance", decreasing = NA)
 #'
 NULL
 
+
```
TuomasBorman: Remove extra line

R/plotSeries.R
```diff
@@ -96,7 +100,9 @@ setGeneric("plotSeries", signature = c("object"),
                     x,
                     y = NULL,
                     rank = NULL,
+                    colour.by = colour_by,
                     colour_by = NULL,
+                    size.by = size_by,
                     size_by = NULL,
                     linetype_by = NULL,
```
TuomasBorman, line 110: linetype_by --> linetype.by

TuomasBorman:
I checked the unit tests. There are many tests that test internal functions, and parameter names of internal functions are not changed if they cannot be used by user. Unit tests are updated and should work after deprecated parameter names are removed in the future.

TuomasBorman (approved):

tests/testthat/test-1treeData.R
```diff
@@ -49,16 +49,13 @@ test_that("tree data", {
...
+    expect_equal( rowTreeData(x, tree.name = "phylo.1"), NULL )
+    expect_error( colTreeData(x, tree.name = 1) )
+    expect_error( rowTreeData(x, tree.name = TRUE) )
+    expect_error( rowTreeData(x, tree.name = c("test", "test2")) )
     td <- rowTreeData(x)
     rowTreeData(x, "test") <- td
     expect_equal(names(x@rowTree), c("phylo", "test"))
-    data(esophagus)
-    tse <- mia::mergeSEs(esophagus, GlobalPatterns)
-    expect_equal( rowTreeData(tse, "phylo"), rowTreeData(esophagus, "phylo") )
```
TuomasBorman, line 63: These tests were removed since mergeSEs works differently since updated. Now it merges the tree. These tests assume that the two trees are stored in output of mergeSEs untouched

Issue #92, opened 2023-09-07
plotRDA, plot loadings


TuomasBorman:
In OMA, there is an example on how to [plot loadings of features in  dbRDA](https://microbiome.github.io/OMA/community-similarity.html#dbrda-workflow). We could consider to generalize this to plotting function if that is a common way to analyze which features drive the difference

ElySeraidarian:
I would like to provide some examples on how to plot these feature loadings after talking with @antagomir about this.

There would be 2 ways of plotting : 

This requires the latest github version of mia.

1) With tree, this option means that the tree need to be agglomerated to the same features as the loadings matrix we get after performing the reduction method (this is now possible for LDA & NMF methods as agglomerateByPrevalence has been fixed to be possible to agglomerate tree). Here are quick examples for each of these reduction methods : 
- PCA 
``` 
library(mia)
library(ggtree)
library(scater)
data("GlobalPatterns", package = "mia")
tse <- GlobalPatterns
tse <- logNormCounts(tse)
# Agglomerate to keep only high prevalence features
tse <- agglomerateByPrevalence(tse, rank="Phylum", prevalence=0.99, update.tree = TRUE)
# Achieve PCA reduction
tse <- runPCA(tse, name = "PCA", ncomponents = 2)
# Get the feature loadings
loadings_matrix <- attr(reducedDim(tse, "PCA"), "rotation")
phylo <- rowTree(tse)
circ <- ggtree(phylo, layout = "circular")
df <- rowData(tse)
color <- randomcoloR::distinctColorPalette(
  length(
    unique(
      df$Phylum
    )
  )
)
df <- data.frame(Class = df$Phylum)
rownames(df) <- phylo$tip.label

df2 <- data.frame(abs(loadings_matrix))
rownames(df2) <- phylo$tip.label

# Plot the tree with first inner circle (Classes)
p <- gheatmap(
  p = circ,
  data = df,
  offset = -.1,
  width = .1,
  colnames_angle = 95,
  colnames_offset_y = .5,
  font.size = 4,
  color = "black") +
  ggplot2::scale_fill_manual(
    values = color,
    name = "Class"
  )
# Plot the feature loadings in different circles
for(i in 1:2){
  if(i == 1){
    p <- p +
      ggnewscale::new_scale_fill()
  }
  df3 <- dplyr::select(
    df2, (all_of(i))
  )

  p <- gheatmap(
    p,
    df3,
    offset = i*.065,
    width = .1,
[...]

PR #135, opened 2024-07-09
Plot feature loadings function

ElySeraidarian:
According to #92, we need a way to plot feature loadings after performing a reduction method by several ways. Here is a first version with PCA that plots feature loadings in 4 different ways : Tree, Barplot, Screeplot & Heatmap. This will be available for LDA & NMF as soon as it is possible to retrieve loadings matrix from reducedDim slot. There are parameters that can be added so that we can talk about it (for example the number of features plotted). The parameter method is probably doomed to disappear as we will be able to recognize which method has been used in the reducedDim slot. There are examples for NMF & LDA at the end for now without using the function. This will need examples in OMA and strong documentation in order to tell people how to use this function (agglomerate tree if plotting the tree, only plotting few features, ...). It is currently possible to give either loadings matrix or TSE object to plot.

I will add tests & generate documentation also later.


R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
+#' Plot tree with feature loadings after performing a reducedDim (e.g. PCA, LDA, NMF)
```
antagomir: Remove " (e.g. PCA, LDA, NMF)" -> that cannot be on the title line. This can be moved to line 4

antagomir: This method works also without tree -> remove "tree"?

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#'
+#' Based on the \code{\link[diffTop:plotASVcircular]{plotASVcircular}} method using phyloseq
```
antagomir: "Based on" -> "Inspired by"

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
+#' Plot tree with feature loadings after performing a reducedDim (e.g. PCA, LDA, NMF)
+#'
+#' Based on the \code{\link[diffTop:plotASVcircular]{plotASVcircular}} method using phyloseq 
+#' and has been converted to use TreeSummarizedExperiment objects.
+#'
+#' @param object a
```
antagomir: "x" would be more standard here (instead of object)

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'   object.
+#' 
+#' @param method One dimensionality reduction method of \code{c("PCA", "LDA", "NMF")}
```
antagomir: Do we need this if the method should also support general numeric loading matrix?

At least this should be optional. Ideally, the method is agnostic to which method generated the loadings.

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#'   indicating which method will be used.
+#'   (default: \code{method = "PCA"})
+#'  
+#' @param rank a single character defining a taxonomic rank. Must be a value of
```
antagomir: Btw is it possible to show tree and PC components with higher resolution than the colored rank info?

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#'   \code{taxonomyRanks()} function.
+#'   
+#' @param plot One way to plot feature loadings of \code{c("Heatmap", "Barplot", "Screeplot", "Tree")}
```
antagomir: I suggest to use here argument name "method"

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#'   
+#' @param plot One way to plot feature loadings of \code{c("Heatmap", "Barplot", "Screeplot", "Tree")} 
+#'   (default: \code{plot = "Heatmap"})
+#'   
+#' @details
+#' # TO DO 
+#' @return 
+#' A \code{ggplot2} object. A circular plot annotated with TreeSummarizedExperiment object.
+#'
+#' @name plotFeatureLoadings
```
antagomir: I think the @name tag is not needed (?). 

The function name could be shorter, just plotLoadings?

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' @examples
+#' 
+#' 
+#' # Plotting feature loadings with tree
+#' library(mia)
+#' library(ggtree)
+#' library(scater)
+#' data("GlobalPatterns", package = "mia")
+#' tse <- GlobalPatterns
+#' tse <- logNormCounts(tse)
```
antagomir: this should use instead the transformAssay function

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' tse <- agglomerateByPrevalence(tse, rank="Phylum", prevalence=0.99, update.tree = TRUE)
```
antagomir: Any chance to keep lower prevalence threshold? 0.99 is very high.

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' tse <- runPCA(tse, name = "PCA", ncomponents = 2)
```
antagomir: Is it possible to have support for more than 2 PC components in the plotting method?

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' plotFeatureLoadings(tse, method= "PCA", plot = "Tree")
```
antagomir: I would use lowercase versions for the method names: "tree", "barplot" etc

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' 
+#' # Plotting without tree as a heatmap
+#' library(mia)
+#' library(ggtree)
+#' library(scater)
+#' data("GlobalPatterns", package = "mia")
+#' tse <- GlobalPatterns
+#' tse <- logNormCounts(tse)
+#' tse <- agglomerateByPrevalence(tse, rank="Phylum", prevalence=0.99, update.tree = TRUE)
+#' tse <- runPCA(tse, name = "PCA", ncomponents = 2)
```
antagomir: No need to repeat these steps a second time in the example

antagomir: (same comment below)

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' #' # Plotting without tree as a screeplot
+#' library(mia)
+#' library(ggtree)
+#' library(scater)
+#' data("GlobalPatterns", package = "mia")
+#' tse <- GlobalPatterns
+#' tse <- logNormCounts(tse)
+#' tse <- agglomerateByPrevalence(tse, rank="Phylum", prevalence=0.99, update.tree = TRUE)
+#' tse <- runPCA(tse, name = "PCA", ncomponents = 2)
+#' plotFeatureLoadings(loadings_matrix, method= "PCA", plot = "Screeplot")
```
antagomir: How about p + coord_flip() as default for better readability of the taxa names?

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' # Plotting without tree as a barplot
+#' library(mia)
+#' library(ggtree)
+#' library(scater)
+#' data("GlobalPatterns", package = "mia")
+#' tse <- GlobalPatterns
+#' tse <- logNormCounts(tse)
+#' tse <- agglomerateByPrevalence(tse, rank="Phylum", prevalence=0.99, update.tree = TRUE)
+#' tse <- runPCA(tse, name = "PCA", ncomponents = 2)
+#' plotFeatureLoadings(loadings_matrix, method= "PCA", plot = "Barplot")
```
antagomir: Barplot could be sorted from the smallest to largest value and the names do not have to match between the two PC components.

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' # Plotting without tree as a heatmap
+#' library(mia)
+#' library(ggtree)
+#' library(scater)
+#' data("GlobalPatterns", package = "mia")
+#' tse <- GlobalPatterns
+#' tse <- logNormCounts(tse)
+#' tse <- agglomerateByPrevalence(tse, rank="Phylum", prevalence=0.99, update.tree = TRUE)
+#' tse <- runPCA(tse, name = "PCA", ncomponents = 2)
+#' plotFeatureLoadings(loadings_matrix, method= "PCA", plot = "Heatmap")
```
antagomir: Could we have just one color legend common to all PC components?

Either a single heatmap, or if heatmaps are separate (per component) then it might work better to sort the values from the smallest to the largest within each PC component.

The color order should be reverted (blue negative, red positive)

antagomir: Title should be PC1, PC2 etc. (not "Feature loadings plot")

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+  if ( !(method %in% c("PCA", "LDA", "NMF")) ) {
+    stop("'method' must be one of c('PCA', 'LDA', 'NMF').", call. = FALSE)
+  }
+  
+  if ( !(plot %in% c("Screeplot", "Barplot", "Tree", "Heatmap")) ) {
+    stop("'method' must be one of c('Screeplot', 'Barplot', 'Tree', 'Heatmap').", call. = FALSE)
+  }
+  
+  if (is.matrix(object) && plot == "Tree") {
+    stop("Tree cannot be plotted for a non-TSE object.", call. = FALSE)
```
antagomir: Write positively: "TreeSummarizedExperiment object is required for the tree plotting".

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+      geom_bar(stat="identity") + xlim(-1,1) + coord_flip()
+    p2 <- ggplot(df, aes(x = .data[["PC2"]], .data[["Feature"]])) +
+      geom_bar(stat="identity") + xlim(-1,1) + coord_flip()
+    p1 + p2
+    
+  }
+  
+  
+}
+##############################################
```
antagomir: Are the lines from 257 onwards necessary?

antagomir (changes requested):
Great! See the comments.

R/plotFeatureLoadings.R
TuomasBorman: `method` could be `dimred`. You can check plotRDA which plots the principal coordinates

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+#' @rdname plotFeatureLoadings
+setGeneric("plotFeatureLoadings", signature = c("object"),
+           function(object, ...)
+             standardGeneric("plotFeatureLoadings"))
+
+
+#' @rdname plotFeatureLoadings
+#' @export 
+setMethod("plotFeatureLoadings", signature = c(object = "TreeSummarizedExperiment"),
+          function(object,
```
TuomasBorman: General comment:

We have agreed to develop the methods strictly following the Bioconductor guidelines. Run BiocCheck::BiocCheck().

The indentation should always be multiplication of 4 spaces. Also the line should have max 80 characters.

Important thing is to add explanations as comments what the code does. This makes the maintenance lots of easier when we know what was the purpose.

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+              )
+              df <- data.frame(Class = df$Phylum)
+              rownames(df) <- phylo$tip.label
+              
+              
+              df2 <- data.frame(abs(loadings_matrix))
+              rownames(df2) <- phylo$tip.label
+              
+              
+              p <- gheatmap(
```
TuomasBorman: Is this gheatmap new dependency?

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+            }
+            
+            if (plot == "Tree") {
+              
+              phylo <- rowTree(object)
+              circ <- ggtree(phylo, layout = "circular")
+              df <- rowData(object)
+              color <- randomcoloR::distinctColorPalette(
+                length(
+                  unique(
```
TuomasBorman: Try to restructure the code. The "main" function should be simple and call internal functions. Here:

1. You could first call `.get_loadings_plot_data` that prepares data.frame from the data
2. The you could call `.loadings_plotter` That creates the plot.

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+              df <- df[ order(df[["PC1"]]), ]
+              df <- df[ order(abs(df[["PC2"]]), decreasing = TRUE)[1:10], ]
+              df <- df[ order(df[["PC2"]]), ]
+            } else {
+              stop("These methods are yet to be implemented", call. = FALSE)
+            }
+            
+            if (plot == "Tree") {
+              
+              phylo <- rowTree(object)
```
TuomasBorman: What if rowTree() is NULL?

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+              for(i in 1:2){
+                if(i == 1){
+                  p <- p +
+                    ggnewscale::new_scale_fill()
+                }
+                df3 <- dplyr::select(
+                  df2, (all_of(i))
+                )
+                
+                p <- gheatmap(
```
TuomasBorman: INstead of using new dependency, you could check https://github.com/microbiome/miaViz/blob/dde325321015e657ff2cf904c9feb8387c9eb2f4/R/plotPrevalence.R#L576

R/plotFeatureLoadings.R
TuomasBorman: `layout` is use in other functions

R/plotFeatureLoadings.R
```diff
@@ -0,0 +1,404 @@
...
+                  low = "gray98",
+                  color = "black",
+                  legend_title = expression(beta[k]))
+              }
+              
+              p
+              
+            } else {
+              .plot_pca_feature_loadings(df, plot)
+            }
```
TuomasBorman: It is hard to see when the function ends. I suggest to save the plot to variable and always end the function with return()

R/plotFeatureLoadings.R
TuomasBorman: For tree plotting, we have this function https://github.com/microbiome/miaViz/blob/dde325321015e657ff2cf904c9feb8387c9eb2f4/R/plotTree.R#L1031 Try to check if you can use that code.

For maintenance, it is always easier if we can use and improve the existing code rather than having multiple implementations on the same thing (tree plotting might not be relevant here though, just a general note. However, for heatmap the existing code might work)

TuomasBorman (changes requested):

R/plotFeatureLoadings.R
ElySeraidarian: this is checked in .check_parameters before going there

TuomasBorman: OK, good!

R/plotFeatureLoadings.R
ElySeraidarian: gheatmap is from ggtree package that is used in others miaViz functions (e.g. plotTree)

R/plotFeatureLoadings.R
ElySeraidarian: For now it doesn't recognize the reduction method used but it will probably be removed once it is done

antagomir: Ok - can we remove this and resolve the conversation, or is this needed?

TuomasBorman: One problem is that in practice the ordination result can be named abitrarily (it can be "pca" or "pca_something"). There should be a way to determine the fetched ordination from reducedDim.

The result should be calculated with runPCA or addLDA, so the result is standardized. If the loadings are not in certain place, you can just give error.

See plotRDA

R/plotFeatureLoadings.R
ElySeraidarian: Yes indeed, it was for my examples to keep very few features

R/plotFeatureLoadings.R
antagomir: Yes. 

And loadings are always a numeric matrix. If we can support arbitrary numeric matrices in addition to TreeSE slot here then that would be worth having I think.

R/plotFeatureLoadings.R
antagomir: How about just aggregating to even higher level? Then you will have less features without need to add prevalence filtering.

R/plotFeatureLoadings.R
ElySeraidarian: Yes I can do it

ElySeraidarian:
Made some changes, I'll add a parameter to increase the number of principal components and maybe one optional parameter to select the name used for the matrix in reducedDim slot if it has been changed by the user? So that I'll be able to recognize the method used

R/plotFeatureLoadings.R
ElySeraidarian: I tried but then it's just a barplot? I don't know how I could increase readability otherwise

R/plotFeatureLoadings.R
ElySeraidarian: I think this should be possible

R/plotFeatureLoadings.R
antagomir: Hmm the difference to barplot is that the bars are sorted here?

ElySeraidarian: They should be sorted in both now

ElySeraidarian:
I have added support for more PC components (I have not added a parameter, it plots the number of components given by the matrix but maybe I should add if user wants to plot less?).  Also added support if name is changed in reducedDim slot.

antagomir:
Yes, it is necessary to give the number of components. Argument name "ncomponents" has been used elsewhere. 

By default it could be ncomponents=5?

R/plotFeatureLoadings.R
TuomasBorman: Thanks, then it is ok

R/plotLoadings.R
```diff
@@ -0,0 +1,341 @@
...
+        }
+    return(p)
+    }
+)
+
+.check_parameters <- function(x, dimred, layout, n,...) {
+    # Checking if dimred is correct
+    if ( !(dimred %in% c("PCA", "LDA", "NMF")) ) {
+        stop("'dimred' must be one of c('PCA', 'LDA', 'NMF').", call. = FALSE)
+    }
```
TuomasBorman: Now user cannot plot reducedDim(tse, "My_PCA")

--> Check this https://github.com/microbiome/miaViz/blob/c1c277a36b8f8ec195cd24544c4fc889ec5dc751/R/plotCCA.R#L313

Update also descrption

TuomasBorman: Also check that teh value is character scalar .is_a_string

R/plotLoadings.R
```diff
@@ -0,0 +1,341 @@
...
+    # Checking if layout is correct
+    if ( !(layout %in% c("screeplot", "barplot", "tree", "heatmap")) ) {
```
TuomasBorman: This fails if user inouts a vector. --> .is_a_string()

R/plotLoadings.R
```diff
@@ -0,0 +1,341 @@
...
+    # Making sure the user doesn't try to plot the tree if he gives only the matrix
+    if (is.matrix(x) && layout == "tree") {
+        stop("TreeSummarizedExperiment object is required for the tree plotting.", call. = FALSE)
+    }
+    # Making sure the tree is not null
+    if (is(x, "TreeSummarizedExperiment")) {
+        if ( is.null(rowTree(x)) && layout == "tree") {
+            stop ("Tree is null.", call. = FALSE)
+        }
+    }
```
TuomasBorman: Try to avoid nested ifs. Moreover, I think this does not catch if the object is SummariedExperiment.

```
if( layout == "tree" && !(is(x, "TreeSummarizedExperiment") && !is.null(rowTree(x, tree.name)) ){

}
```

TuomasBorman: Moreover, user should be able to specify tree.name (the name of rowTree) --> check other methods

R/plotLoadings.R
```diff
@@ -0,0 +1,341 @@
...
+
+}
+
+.get_loadings_plot_data <- function(x, n, ncomponents) {
+    L <- list()
+    for (i in 1:ncomponents) {
+        df <- data.frame(x)
+        df[["Feature"]] <- rownames(x)
+        # Ordering loadings
+        df <- df[ order(abs(df[[paste("PC",as.character(i),sep="")]])[1:n], decreasing = TRUE), ]
```
TuomasBorman: I believe you can do this without loop. (Loops should be avoided in R if possible, they are slow. Also good practice to avoid loops)

Moreover, you are handling the data with column names. However, it is not necessary since you know the index of columns. It is preferred to use indices since that is more general approach (column names can differ; they are not always "PC*")

R/plotLoadings.R
```diff
@@ -0,0 +1,341 @@
...
+        df <- df[ order(df[[paste("PC",as.character(i),sep="")]]), ]
+        # Add factor to keep order
+        df[["Feature"]] <- factor(df[["Feature"]], levels = df[["Feature"]])
+        L <- append(L,df)
+    }
+    return(L)
+}
+
+.loadings_tree_plotter <- function(x, loadings_matrix, ncomponents) {
+    phylo <- rowTree(x)
```
TuomasBorman: User should be able to specify the tree. --> tree.name parameter

TuomasBorman: Default value "phylo". Moreover, then you should subset the data based on the tree (TreeSE can have multiple trees, and tree might be missing some rows).

tree <- rowTree(tse, tree.name)
ind <- rowLinks(tse)[["whichTree"]] == tree.name
if( any(ind) ){
warning("Data was subsettd")
}
tse_sub <- tse[ind, ]

R/plotLoadings.R
```diff
@@ -0,0 +1,341 @@
...
+        for(i in 1:ncomponents){
+            if(i == 1){
+                p <- p +
+                ggnewscale::new_scale_fill()
+            }
+            df3 <- dplyr::select(
+                df2, (all_of(i))
+            )
+        
+            p <- gheatmap(
```
TuomasBorman: Add comments

TuomasBorman (changes requested):

R/plotLoadings.R
ElySeraidarian: This is because user can decide to give only the loadings matrix in this case it will return error because there is no rowTree in a matrix

ElySeraidarian: Nevermind seems to work, I believed it would go through all the checks in if statement but it stops if first one is incorrect

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+#' Inspired by the \code{\link[diffTop:plotASVcircular]{plotASVcircular}} method using phyloseq 
+#' and has been converted to use TreeSummarizedExperiment objects.
+#'
+#' @param x a
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'   x.
+#' 
+#' @param dimred One dimensionality reduction method of reducedDimNames
+#'   indicating which method will be used.
+#'   (default: \code{dimred = "PCA"})
```
TuomasBorman: Put this parameter descriptions in this format.

Class. Definition. Default value.

-->

dimred \code{Character scalar}. DImension reduction from reducedDimNames to be plotted. (Default: \code{"PCA"])

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+    }
+    #Checking if n is a positive number
+    if ( !(is.numeric(n) && n > 0) ) {
+        stop("'n' must be a positive number.", call. = FALSE)
+    }
+    #Checking if ncomponents is a positive number 
+    if ( !(is.numeric(ncomponents) && ncomponents > 0) ) {
+        stop("'ncomponents' must be a positive number.", call. = FALSE)
+    }
+
```
TuomasBorman: R functions always return something. I think it is good practice to specify the returned value even though the value is not used. it could be just return(NULL)

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+    }
+    # Making sure the user doesn't try to plot the tree if he gives only the matrix
+    if (is.matrix(x) && layout == "tree") {
+        stop("TreeSummarizedExperiment object is required for the tree plotting.", call. = FALSE)
+    }
+    # Making sure the tree is not null
+    if( layout == "tree" && !(is(x, "TreeSummarizedExperiment") && !is.null(rowTree(x, tree_name)) )) {
+        stop ("Tree is null.", call. = FALSE)
+    }
+    #Checking if n is a positive number
```
TuomasBorman: Add space after #

TuomasBorman: Check all the comments and add spaces

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+}
+
+.check_components <- function(x, ncomponents) {
+    #Checking if there are enough components in the matrix
+    if (ncomponents > length(colnames(x))) {
+        stop("'ncomponents' must be lower or equal than number of components.", call. = FALSE)  
+    }
+}
```
TuomasBorman: length(colnames()) -> ncol()

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+            df <- data.frame(PC =L[[i]][[i]],
+                Feature = L[[i]][["Feature"]])
+            p <- p + ggplot(df, aes(x = .data[["PC"]], .data[["Feature"]])) +
+                geom_bar(stat="identity") +
+                theme(axis.title.x = element_blank()) +
+                xlim(-1,1) +
+                labs(title=paste("PC",i,sep=""))
+        }
+
+        return(p)
```
TuomasBorman: I think it is more clear that the function returns values in the end of the function. You could move return(p) to end , and remove them from inside if statements

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+#' @return 
+#' A \code{ggplot2} object. A circular plot annotated with TreeSummarizedExperiment object.
+#'
+#' @name plotLoadings
+#' @export
+#' 
+#' @author
+#' Ely Seraidarian
+#' Contact: \url{microbiome.github.io}
+#' @examples
```
TuomasBorman, line 47: Add line break above examples

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+#'
```
TuomasBorman, line 48: remove extra line

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+        # Making sure there is no error in parameters given by the user
+        .check_parameters(x,
+                        dimred = dimred,
+                        layout = layout,
+                        n = n,
+                        ncomponents = ncomponents,
+                        tree_name = tree_name,
+                        ...)  
+
+        loadings_matrix <- attr(reducedDim(x, dimred), "rotation")
```
TuomasBorman: This works now with PCA. How about with RDA (mia::getRDA)

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+            ...) {
+        
+            
+        # Making sure there is no error in parameters given by the user
+        .check_parameters(x,
+                        dimred = dimred,
+                        layout = layout,
+                        n = n,
+                        ncomponents = ncomponents,
+                        tree_name = tree_name,
```
TuomasBorman: tree_name --> tree.name

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+#' @param layout One way to plot feature loadings of \code{c("heatmap", "barplot", "screeplot", "tree")} 
+#'   (default: \code{layout = "heatmap"})
+#' 
+#' @param n A numeric specifying the number of features to be plotted.
+#'   (default: \code{n = 10})
+#'   
+#' @param ncomponents A numeric specifying the number of components.
+#'   (default: \code{ncomponents = 5})
+#' 
+#' @param tree_name a single \code{character} value specifying a rowTree/colTree from
```
TuomasBorman: This cannot specify colTree

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+                        n = n,
+                        ncomponents = ncomponents,
+                        tree_name = tree_name,
+                        ...)  
+
+        loadings_matrix <- attr(reducedDim(x, dimred), "rotation")
+        #Checking if there are enough components in the matrix
+        .check_components(loadings_matrix, ncomponents)
+        # Ordering loadings and adding factor to keep the order
+        L <- .get_loadings_plot_data(loadings_matrix, n, ncomponents)
```
TuomasBorman: THis is not used if layout == "tree". You can move it inside` else`

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+)
+#' @rdname plotLoadings
+#' @export 
+setMethod("plotLoadings", signature = c(x = "matrix"),
+    function(x,
+            dimred = "PCA",
+            layout = "heatmap",
+            n = 10,
+            ncomponents = 5,
+            tree_name = "phylo",
```
TuomasBorman: tree_name is not used in this function?

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+    df[["Feature"]] <- factor(df[["Feature"]], levels = df[["Feature"]])
+    return(df)
+}
+
+.get_loadings_plot_data <- function(x, n, ncomponents) {
+  # Transform into a dataframe
+  df <- data.frame(x)
+  # Keep only the number of components needed
+  names <- colnames(df)[1:ncomponents]
+  df <- dplyr::select(df, all_of(names))
```
TuomasBorman: You could just perhaps:

df <- df[, seq_len(ncomponents), drop = FALSE]

R/plotLoadings.R
```diff
@@ -0,0 +1,374 @@
...
+    # Ordering by actual loadings
+    df <- df[order(df[[i]]), ]
+    # Add factor to keep order
+    df[["Feature"]] <- factor(df[["Feature"]], levels = df[["Feature"]])
+    return(df)
+}
+
+.get_loadings_plot_data <- function(x, n, ncomponents) {
+  # Transform into a dataframe
+  df <- data.frame(x)
```
TuomasBorman: as.data.frame()

TuomasBorman: To save memory, it might be better to store the data back to x

TuomasBorman (changes requested):
Couple comments

ElySeraidarian:
Hadn't seen comments I'll check this

R/plotLoadings.R
ElySeraidarian: This does not work for now, but @thpralas will create a wrapper for every reducedDim methods and store in feature_loadings instead of rotation so it will work for every method. I tried and currently RDA loadings matrix is not stored in rotation slot so it will not work here

R/plotLoadings.R
ElySeraidarian: I tried and somehow it wasn't working so I went back to previous version (maybe I made a mistake somewhere)

R/plotLoadings.R
TuomasBorman: Wrappers might not be optimal solutions since it requires more maintenance and it might be that users still use scater::runPCA instead of our wrapper. It might be then quite confusing for user why the function does not work even though user used the "correct" function.

Instead, you could create an internal function

```
loading_names <- c("rotation", "loadings", "some_other_name")
attr_names <- names(attributes(reduced_dim))
attr_name <- attr_names[ attr_names %in% loading_names ]
if( length(attr_name) != 1 )){
    stop("Loadings cannot be found..")
}
loadings <- attr(tse, attr_name)
```

TuomasBorman: The code does not necessarily work but gives the idea

ElySeraidarian: I see but for example for RDA there is no attribute where it is stored, so you can't even retrieve the matrix that easily

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+#'
+#' Inspired by the \code{\link[diffTop:plotASVcircular]{plotASVcircular}} method using phyloseq 
+#' and has been converted to use TreeSummarizedExperiment objects.
+#'
+#' @param x a
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'   x.
+#' 
+#' @param dimred \code{Character scalar}. Dimension reduction from reducedDimNames to be plotted. 
+#'   (default: \code{dimred = "PCA"})
```
TuomasBorman: (Default: \code{"PCA"}) is simpler. Other functions follow that style

TuomasBorman: Modify also other parameter descriptions

R/plotLoadings.R
TuomasBorman: Sorry, last week I was busy with OMA, so I had not time to check this.

From RDA, you can get "constrained weighted orthonormal species scores" which are analogous to loadings of PCA. See for instance this https://www.rdocumentation.org/packages/vegan/versions/2.4-2/topics/cca.object

When used getRDA, (getCCA) or add* functions, these are stored into "rotation" attribute of returned matrix.

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+}
+
+
+.loadings_tree_plotter <- function(x, loadings_matrix, ncomponents, tree.name) {
+    # Retrieve rowTree
+    phylo <- rowTree(x, tree.name)
+    # Subset data based on the tree
+    ind <- rowLinks(x)[["whichTree"]] == tree.name
+    if( any(ind) ){
+      warning("Data was subsetted")
```
TuomasBorman: warning("message here", call. = FALSE)

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+        }
+    }
+    
+    else if (layout == "screeplot") {
+        # Transform into a dataframe
+        df <- data.frame(PC = L[[1]][[1]], Feature = L[[1]][["Feature"]]) 
+        # Plot first component
+        p <- ggplot(df, aes(x = .data[["PC"]], .data[["Feature"]])) +
+            geom_bar(stat="identity") +
+            theme(axis.title.x = element_blank(),
```
TuomasBorman: This is not scree plot. Scree plot includes eigenvalues

TuomasBorman: As this is same as bar plot, this can be removed

TuomasBorman:
For heatmap, "shadow text" might be better

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+
+#' @rdname plotLoadings
+setGeneric("plotLoadings", signature = c("x"),
+    function(x, ...) 
+        standardGeneric("plotLoadings"))
+
+
+#' @rdname plotLoadings
+#' @export 
+setMethod("plotLoadings", signature = c(x = "TreeSummarizedExperiment"),
```
TuomasBorman, line 85: We could enable the function for SingleCellExperiment objects also but then missing rowTree slot should be taken into acount (SCE objects do not have trees)

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+  # Keep only the number of components needed
+  names <- colnames(x)[1:ncomponents]
+  df <- dplyr::select(x, all_of(names))
+  # Add feature labels
+  x[["Feature"]] <- rownames(x)
+  # Apply the function to each component and return the list
+  L <- lapply(1:ncomponents, .process_component, df = x, n = n)
+  return(L)
+}
+
```
TuomasBorman, line 310: Remove extra line break

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+    ind <- rowLinks(x)[["whichTree"]] == tree.name
+    if( any(ind) ){
+      warning("Data was subsetted")
+    }
+    x <- x[ind, ]
+    # Store plot tree
+    circ <- ggtree::ggtree(phylo, layout = "circular")
+    df <- rowData(x)
+    # Transform into a dataframe
+    df <- data.frame(Class = df$Phylum)
```
TuomasBorman: User should be able to choose the rank (or either no rank at all, i.e., rownames)

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+    # Match labels
+    rownames(df) <- phylo$tip.label
```
TuomasBorman: This can cause serious problems. That is because rows

1. do not have to match with tips
2. The order might not be the same in rows and nodes of tree

You can get the linkage between rows and nodes of tree from rowLinks()

TuomasBorman: The point 1 can cause problems even after rowLinks() since your method is assuming that rows are in tips of tree?

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+    df[["Feature"]] <- factor(df[["Feature"]], levels = df[["Feature"]])
+    return(df)
+}
+
+.get_loadings_plot_data <- function(x, n, ncomponents) {
+  # Transform into a dataframe
+  x <- as.data.frame(x)
+  # Keep only the number of components needed
+  names <- colnames(x)[1:ncomponents]
+  df <- dplyr::select(x, all_of(names))
```
TuomasBorman: Instead of dplyr::select(), use importFrom above the function

@importFrom dplyr select

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+    # Retrieve rowTree
+    phylo <- rowTree(x, tree.name)
+    # Subset data based on the tree
+    ind <- rowLinks(x)[["whichTree"]] == tree.name
+    if( any(ind) ){
+      warning("Data was subsetted")
+    }
+    x <- x[ind, ]
+    # Store plot tree
+    circ <- ggtree::ggtree(phylo, layout = "circular")
```
TuomasBorman: Here also. Add

#' @importFrom ggtree ggtree

above

.loadings_tree_plotter

R/plotLoadings.R
TuomasBorman: Although, you could check whether the rows are in tips, and if they are not, then you could give informative error

TuomasBorman: Then if the tree is not agglomerated in prior to this function, you could prune it with mia::.prune_tree

R/plotLoadings.R
```diff
@@ -0,0 +1,362 @@
...
+                Feature = L[[i]][["Feature"]])
+            p <- p + ggplot(df, aes(x = .data[["PC"]], .data[["Feature"]])) +
+                geom_bar(stat="identity") +
+                theme(axis.title.x = element_blank(),
+                    axis.text.x = element_text(angle = 90,
+                        vjust = 0.5, hjust=1, size=6)) +
+                xlim(-1,1) +
+                labs(title=names[i]) +
+                coord_flip()
+        }
```
TuomasBorman: Instead of looping like this, you could use facets which is more neat solution (then all the data must be in same data.frame)

TuomasBorman:
```
# Prepare the data in correct format, this might not take into account
# the correct order and features, but it gives the idea
df <- lapply(seq_len(length(L)), function(i){
    L[[i]][, i, drop = FALSE]
})
df <- do.call(cbind, df) 
df[["feat"]] <- rownames(df)

# To long format
df <- reshape(
    df, varying =colnames(df)[ !colnames(df) %in% c("feat") ],
    v.names = "value", 
    timevar = "PC",
    times = colnames(df)[ !colnames(df) %in% c("feat") ],
    direction = "long")

ggplot(df, aes(x = .data[["value"]], y = .data[["feat"]])) +
    geom_bar(stat="identity") +
    facet_wrap(~ PC, scales = "free")

```

TuomasBorman: Check also other plot types. I think the data.frame list might not be needed. Just use facets instead of looping over list of data.frames

ElySeraidarian: This is indeed a nice solution but after trying for hours, haven't been able to make axes label change for every component like it used to be. Also ordering seems to be working only on first component for some reason. 
image
Here is the best I have been able to do but I guess it looks worse

antagomir: I think you would need to define every (sub)figure separately in order to have a full control.

ElySeraidarian:
Every change should have been done according to your comments, the values in tree layout is now accurately matching with labels (getting same results with heatmap and barplot). I'll focus on making it operational for every reduction method (detecting which method has been performed) with what @thpralas has done ( and maybe find a solution if user is not aware of these wrappers)

antagomir:
Just for clarity, could you "resolve" all comments that have been addressed?

antagomir:
Solution for non-wrapper users would be to support plain numeric matrices (without tree) in addition to (Tree)SE objects. If users want to include a tree then I think it is better to require that they create a TreeSE object.

R/plotLoadings.R
```diff
@@ -0,0 +1,385 @@
+#' Plot feature loadings after performing a reducedDim 
+#'
+#' Inspired by the \code{\link[diffTop:plotASVcircular]{plotASVcircular}} method using phyloseq 
+#' and has been converted to use TreeSummarizedExperiment objects.
```
antagomir: I would here include one sentence description of what this function does.

And then the current lines 3-4 could go to @details section.

R/plotLoadings.R
```diff
@@ -0,0 +1,385 @@
+#' Plot feature loadings after performing a reducedDim
```
antagomir: Change this to: "Plot feature loadings for TreeSummarizedExperiment objects."

Then later in @details explain that the TreeSE object is expected to have content in reducedDim slot.

However TreeSE inherits SingleCellExperiment, and SingleCellExperiments have reducedDim slot (but no tree). Could we make this more generic and add support for SCE objects?

R/plotLoadings.R
```diff
@@ -0,0 +1,385 @@
+#' Plot feature loadings after performing a reducedDim 
+#'
+#' Inspired by the \code{\link[diffTop:plotASVcircular]{plotASVcircular}} method using phyloseq 
+#' and has been converted to use TreeSummarizedExperiment objects.
+#'
+#' @param x a
```
antagomir: TreeSummarizedExperiment or SingleCellExperiment? (the latter does not have a tree and it is more generic, TreeSE is a SCE).

R/plotLoadings.R
```diff
@@ -0,0 +1,385 @@
...
+#'   (default: \code{5})
+#' 
+#' @param tree.name A single \code{character} value specifying a rowTree from a
+#'   TreeSummarizedExperiment object. (By default: \code{"phylo"})
+#'   
+#' @param class A single \code{character} value specifying a rank from taxonomyRanks
+#'   (default: \code{rownames})
+#'     
+#' 
+#' @details
```
antagomir, line 33: You can also say here something about the limitations related to feature and sample set sizes etc.

R/plotLoadings.R
ElySeraidarian: Yes, should have updated description but I added SCE support in last commits.

ElySeraidarian:
I have been able to make it work for LDA method with the wrapper 
``` 
library(mia)
library(ggtree)
library(scater)
data("GlobalPatterns", package = "mia")
tse <- GlobalPatterns
tse <- transformAssay(tse, method = "clr", pseudocount = 1)
tse <- agglomerateByPrevalence(tse, rank="Phylum", update.tree = TRUE)
tse <- addLDA(tse)
plotLoadings(tse,  ncomponents = 2)
``` 
However I have not been able to autodetect method if reducedDim slot contains several methods, then user will have to specify name using dimred = "method_name" and I don't believe this is possible either?

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+            layout = "tree",
+            n = 10,
+            ncomponents = 5,
+            tree.name = "phylo",
+            class = rownames(x),
+            ...) {
+      
+        if (length(reducedDimNames(x)) == 1) {
+            dimred <- reducedDimNames(x)
+        }
```
TuomasBorman: See other functions that plot reducedDim slot. I think, default value of dimred should not be determined and this line can be removed.

1. The functions should be similar
2. The functions should be transparent. In practice, this does differently than user has specified

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+                          layout = layout,
+                          n = n,
+                          ncomponents = ncomponents,
+                          ...)  
+        
+        loading_names <- c("rotation", "loadings")
+        attr_names <- names(attributes(reducedDim(x, dimred)))
+        attr_name <- attr_names[ attr_names %in% loading_names ]
+        if( length(attr_name) != 1 ) {
+            stop("Loadings cannot be found..")
```
TuomasBorman: call. = FALSE

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+        stop("'n' must be a positive number.", call. = FALSE)
+    }
+    # Checking if ncomponents is a positive number 
+    if ( !(is.numeric(ncomponents) && ncomponents > 0) ) {
+        stop("'ncomponents' must be a positive number.", call. = FALSE)
+    }
+    # Checking if class is correct
+    if ( is(x, "TreeSummarizedExperiment") && !(any(class %in% taxonomyRanks(x)) || identical(class, rownames(x)))) {
+        stop("'class' must be one of taxonomyRanks", call. = FALSE)
+    }
```
TuomasBorman: We use "rank" in other methods. Should be the same

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+#'   (default: \code{10})
+#'   
+#' @param ncomponents A numeric specifying the number of components.
+#'   (default: \code{5})
+#' 
+#' @param tree.name A single \code{character} value specifying a rowTree from a
+#'   TreeSummarizedExperiment object. (By default: \code{"phylo"})
+#'   
+#' @param class A single \code{character} value specifying a rank from taxonomyRanks
+#'   (default: \code{rownames})
```
TuomasBorman: The parameter name could be "rank". Moreover, the default value could be NULL --> which means that no rank from rowData

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+#'   \code{\link[TreeSummarizedExperiment:TreeSummarizedExperiment-class]{TreeSummarizedExperiment}}
+#'   x.
+#' 
+#' @param dimred \code{Character scalar}. Dimension reduction from reducedDimNames to be plotted. 
+#'   (default: \code{"PCA"})
+#'  
+#' @param layout One way to plot feature loadings of \code{c("heatmap", "barplot", "tree")} 
+#'   (default: \code{"tree"} for TSE, \code{"heatmap"} for others)
+#' 
+#' @param n A numeric specifying the number of features to be plotted.
```
TuomasBorman: Check all the parameter descriptions. They should follow the same converntion that in the other functions

Class. Description. Default

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+                colnames = FALSE) + 
+                scale_fill_gradient2(limits = c(-1,1),
+                    low = "darkslateblue", mid = "white",
+                    high = "darkred", name = "Value")
+            p <- p + labs(title = "Tree feature loadings plot") +
+              theme(legend.key.size = unit(0.5, 'cm'),
+                    plot.title = element_text(size = 18, hjust=0.5))
+        }
+    return(p)
+}
```
TuomasBorman: Legend title is now "Class" but it could be the specified rank

Moreover, usually titles are not added to article figures, so it might be better to not add the title by default (User can easily add it afterwards)

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+    function(x, ...) 
+        standardGeneric("plotLoadings"))
+
+
+#' @rdname plotLoadings
+#' @export 
+setMethod("plotLoadings", signature = c(x = "TreeSummarizedExperiment"),
+    function(x,
+            dimred = "PCA",
+            layout = "tree",
```
TuomasBorman: I would keep the default layout same for TreeSE

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+    cnames <- cnames[!cnames %in% c("Feature") ]
+    
+    new_df <- data.frame(matrix(-1, nrow = length(names), ncol = length(cnames)), row.names = names)
+    colnames(new_df) <- cnames
+    for (i in seq_along(L)) {
+      pc_df <- df[[i]]
+      name <- colnames(pc_df)
+      new_df[rownames(pc_df), name] <- pc_df[, 1]
+    }
+    new_df[["Feature"]] <- names
```
TuomasBorman: Instead of construting the single df in here, I think it might be much neater that the input is already one single data.frame. That helps also the maintenance.

antagomir:
COuld you use `reducedDimNames(tse)` to detect?

R/plotLoadings.R
TuomasBorman: How about this?

```
library(tidytext)
ggplot(df, aes(x = Value, y = reorder_within(Feature, Value, PC))) +
    geom_bar(stat = "identity") +
    scale_y_reordered() +
    facet_wrap(~ PC, scales = "free") +
    theme_minimal() +
    labs(x = "Value", y = "Feature")
```

R/plotLoadings.R
TuomasBorman: Also while tree layout looks nice, it is rather complex.

R/plotLoadings.R
```diff
@@ -0,0 +1,416 @@
...
+        p <- .plot_pca_feature_loadings(L, layout, n, ncomponents)
+        
+        return(p)
+    }
+)
+
+.check_parameters <- function(x, dimred, layout, n, ncomponents, tree.name, class, ...) {
+    # Check tree.name
+    if( is(x, "TreeSummarizedExperiment") && !(tree.name %in% rowTreeNames(x)  && .is_a_string(tree.name))){
+        stop("'tree.name' must be a single character value specifying a colTree.", call. = FALSE)
```
TuomasBorman: rowTree

TuomasBorman: Also this is failing while it should not

```
data("GlobalPatterns", package = "mia")
tse <- GlobalPatterns
tse <- transformAssay(tse, method = "clr", pseudocount = 1)
tse <- agglomerateByPrevalence(tse, rank="Phylum", update.tree = TRUE)
tse <- runPCA(tse, ncomponents = 5, assay.type = "clr")
plotLoadings(tse, layout = "barplot")
```

TuomasBorman: Not all TreeSE have tree

R/plotLoadings.R
TuomasBorman: This works for heatmap
```

library(tidytext)
library(shadowtext)
ggplot(df, aes(x = .data[["PC"]], y = reorder_within(Feature, Value, PC), label = round(.data[["Value"]], 2)))  +
    geom_point(aes(fill = .data[["Value"]]), size = 15-0.4*n, shape = 22) +
    scale_fill_gradient2(
        limits = c(-1,1), low = "darkslateblue", mid = "white", high = "darkred") +
    scale_y_reordered() +
    geom_shadowtext(size = 4-0.075*n) +
    facet_wrap(~ PC, scales = "free")
```

TuomasBorman:
For default layout, I would keep barplot as it is the most simple and fastest to look

TuomasBorman:
> I have been able to make it work for LDA method with the wrapper
> 
> ```
> library(mia)
> library(ggtree)
> library(scater)
> data("GlobalPatterns", package = "mia")
> tse <- GlobalPatterns
> tse <- transformAssay(tse, method = "clr", pseudocount = 1)
> tse <- agglomerateByPrevalence(tse, rank="Phylum", update.tree = TRUE)
> tse <- addLDA(tse)
> plotLoadings(tse,  ncomponents = 2)
> ```
> 
> However I have not been able to autodetect method if reducedDim slot contains several methods, then user will have to specify name using dimred = "method_name" and I don't believe this is possible either?

That is ok in my opinion. Other functions also assume that user specifies the reducedDim to plot.

For instance:

plotReducedDim(tse, dimred = "PCA") --> you have to always specify dimred in this function

TuomasBorman:
I am also wondering if current heatmap layout takes full advantage of heatmaps. Heatmaps are create way to visualize large batch of data. However, in this layout it rather makes the plot more complex compared to simpler barplot. Instead, the heatmap could be in this format.

image


Current heatmap

image

Barplot with facets

image

TuomasBorman:
After thinking this, we could have 2 layouts: "barplot" and "heatmap". The we could have hidden parameter "add.tree" that decides whether to add tree to heatmap (the "tree" layout is also one kind of heatmap)

The implementation could be then simplified a lot without losing features. This helps in the maintenance.

R/plotLoadings.R
ElySeraidarian: > Also this is failing while it should not
> 
> ```
> data("GlobalPatterns", package = "mia")
> tse <- GlobalPatterns
> tse <- transformAssay(tse, method = "clr", pseudocount = 1)
> tse <- agglomerateByPrevalence(tse, rank="Phylum", update.tree = TRUE)
> tse <- runPCA(tse, ncomponents = 5, assay.type = "clr")
> plotLoadings(tse, layout = "barplot")
> ```

I don't see why this would not work? It runs fine for me

TuomasBorman: Forgot to add rowTree(tse) <- NULL

data("GlobalPatterns", package = "mia")
tse <- GlobalPatterns
rowTree(tse) <- NULL
tse <- transformAssay(tse, method = "clr", pseudocount = 1)
tse <- agglomerateByPrevalence(tse, rank="Phylum", update.tree = TRUE)
tse <- runPCA(tse, ncomponents = 5, assay.type = "clr")
plotLoadings(tse, layout = "barplot")

R/plotLoadings.R
ElySeraidarian: Maybe I could still keep it so that user do not have to specify dimred when only one reduction method has been performed (removed default value though)

TuomasBorman: I am not fan of this solution since it is not transparent. we should allow user to get error if user causes it. If we do not do that it is possible source of error. at least it should be

dimred =reducrddimnanmes(tse)[[1]]

In the function definition

ElySeraidarian:
Changes have been made, heatmap has been changed and parameter add.tree instead of tree layout

R/plotLoadings.R
```diff
@@ -0,0 +1,386 @@
...
+    }
+    x <- x[ind, ]
+    # Store plot tree
+    circ <- ggtree(phylo, layout = "circular")
+    df <- rowData(x)
+    # Transform into a dataframe
+    if (is.null(rank)) {
+        df <- data.frame(Class = rownames(x))
+        legend_name <- "Class"
+    } else {
```
TuomasBorman: I think this plotter should take preprocessed data as input. The processing should happen in different internal function

R/plotLoadings.R
```diff
@@ -0,0 +1,386 @@
...
+#' @importFrom dplyr select
+.get_loadings_plot_data <- function(df, n, ncomponents) {
+  # Transform into a dataframe
+  df <- as.data.frame(df)
+  # Keep only the number of components needed
+  df <- df[ , seq_len(ncomponents), drop = FALSE]
+  # Apply the function to each component and return the list
+  res <- lapply(seq_len(ncomponents), .process_component, df = df, n = n)
+  # Combine to single data.frame
+  res <- do.call(rbind, res)
```
TuomasBorman: I simplified this, check that it still works. 

instead of `1:ncomponents`, use `seq_len(ncomponents)`

R/plotLoadings.R
```diff
@@ -0,0 +1,386 @@
...
+                        n = n,
+                        ncomponents = ncomponents,
+                        ...)
+                      
+        # Checking if there are enough components in the matrix
+        .check_components(x, ncomponents)
+        
+        # Plot features with heatmap layout
+        if (layout == "heatmap") {
+            p <- ComplexHeatmap::Heatmap(x, heatmap_legend_param = list(title = "Value"))
```
TuomasBorman: This looks good, but the problem is that all other methods output ggplot objects. --> create a function that gets data from

.get_loadings_plot_data(x, n, ncomponents)

as input and returns heatmap created with ggplot2

TuomasBorman (changes requested):

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: miaViz
 Title: Microbiome Analysis Plotting and Visualization
-Version: 1.13.9
+Version: 1.14.0
```
TuomasBorman: Version should be 1.13.10

1.xx.yy

- xx denotes the Bioc release/devel version (updates every 6 months)
- yy rolling number

I will check this PR tomorrow at latest

TuomasBorman:
I simplified the code a lot. Check that it still works as expected. Also run BiocCheck::BiocCheck() and fix issues related to this added R file (especially, do not exceed 80 characters in line width). Also use inheritParam tag in the documentation.

TuomasBorman:
Thing to discuss: should `rank` be `row.var`? --> variable defining a column from rowData. That is what it is doing currently

ElySeraidarian:
> Thing to discuss: should `rank` be `row.var`? --> variable defining a column from rowData. That is what it is doing currently

Yes I believe this would make more sense

antagomir:
Seems good.

ElySeraidarian:
But we could also still create a parameter rank that agglomerates tree as in the other functions

TuomasBorman:
> But we could also still create a parameter rank that agglomerates tree as in the other functions

If we agglomerate the data, we lose the match between loadings and rows. Loadings are for features of un-agglomerated data --> that is my first impression, check if that is correct

ElySeraidarian:
> > But we could also still create a parameter rank that agglomerates tree as in the other functions
> 
> If we agglomerate the data, we lose the match between loadings and rows. Loadings are for features of un-agglomerated data --> that is my first impression, check if that is correct

Yes that's right, agglomeration has to be done before performing the reduction method

TuomasBorman:
1) I removed authors. See the issue: https://github.com/microbiome/mia/issues/628
2) You should perform first agglomeration and then transform the data (fixed)
3) I added a check that the feature names match in tree heatmap

TuomasBorman (approved):

TuomasBorman:
Looks very nice. Are you interested in updating OMA: https://microbiome.github.io/OMA/docs/devel/pages/beta_diversity.html#sec-dbrda-workflow

If I remember correctly, there were some error?

ElySeraidarian:
> Looks very nice. Are you interested in updating OMA: https://microbiome.github.io/OMA/docs/devel/pages/beta_diversity.html#sec-dbrda-workflow
> 
> If I remember correctly, there were some error?

I will update using this function, I found same results when plotting. Also this example shows how to retrieve the loadings (for now we can plot only by getting matrix and then giving the matrix to plot).

PR #145, opened 2024-07-30
temporal fix rworkflows

0xMuluh:


antagomir (approved):

antagomir:
Checks fail?

0xMuluh:
hmm, it seems the source of the problem is rworkflows. The version of `mia` used is outdated(not devel) with our current changes. I initially assumed it was the `magick` package but it seems not. However I have bumped the required `devel` version for now. In the next release cycle, we can restore the workflow.

0xMuluh:
This is good to go.

PR #148, opened 2024-08-20
add support for groups other than rank in plotAbundance

0xMuluh:
This commit deprecates rank and adds support for other columns in rowData
ping: #100

### DIFF
--- a/R/plotAbundance.R
+++ b/R/plotAbundance.R
@@ -90,9 +90,9 @@
 #' data(GlobalPatterns, package="mia")
 #' tse <- GlobalPatterns
 #' 
-#' ## If rank is set to NULL (default), agglomeration is not done. However, note
-#' ## that there is maximum number of rows that can be plotted. That is why
-#' ## we take sample from the data.
+#' # If rank is set to NULL (default), agglomeration is not done. However, note
+#' # that there is maximum number of rows that can be plotted. That is why
+#' # we take sample from the data.
 #' set.seed(26348)
 #' sample <- sample(rownames(tse), 20)
 #' tse_sub <- tse[sample, ]
@@ -100,12 +100,12 @@
 #' tse_sub <- transformAssay(tse_sub, method = "relabundance")
 #' plotAbundance(tse_sub, assay.type = "relabundance")
 #' 
-#' ## Plotting counts using the first taxonomic rank as default
+#' # Plotting counts using the first taxonomic rank as default
 #' plotAbundance(
 #'     tse, assay.type="counts", group = "Phylum") +
 #'     labs(y="Counts")
 #' 
-#' ## Using "Phylum" as rank. Apply relative transformation to "counts" assay.
+#' # Using "Phylum" as rank. Apply relative transformation to "counts" assay.
 #' plotAbundance(
 #'     tse, assay.type="counts", group = "Phylum", add_legend = FALSE,
 #'     as.relative = TRUE)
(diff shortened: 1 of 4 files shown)


R/plotAbundance.R
```diff
@@ -187,12 +190,15 @@ setMethod("plotAbundance", signature = c("SummarizedExperiment"),
...
                 call. = FALSE)
         }
-        if(!is.null(rank)){
-            .check_taxonomic_rank(rank, x)
+        if(!is.null(group)){
+            if(!(group %in% colnames(rowData(x)))){
+                stop("'group' must be a column from rowData .",
+                     call. = FALSE)
+            }
         }
```
TuomasBorman: Avoid nested ifs, instead use just

`!is.null(group) && !group %in% colnames(rowData(x))`

TuomasBorman:
Thanks, seems to work! However, I think we should think if this is needed as this makes the function more complex. The same can be achieved as follows which might be more transparent (I prefer that)

```
data(GlobalPatterns, package="mia")
tse <- GlobalPatterns

rowData(tse)$bacteria_group <- sample(c("gram-pos", "gram-neg"), nrow(tse), replace = TRUE)

# Works fine
plotAbundance(tse, group = "bacteria_group", as.relative = TRUE)

# The alternative apporach
tse_sub <- agglomerateByVariable(tse, f = "bacteria_group", by = "rows")
tse_sub <- transformAssay(tse_sub, method = "relabundance")
plotAbundance(tse_sub, assay.type = "relabundance")
```
Second thing is that, `group` argument name seems not to be used in anywhere else. It might be good to have same naming as in other functions because that might help users to understand what the arguments do.

Instead of changing the name, one option is to "hiddenly" support other grouping variables. The name could be `rank`, but it would still work with other variables also. As grouping by rank is the most common thing, that might make the function more simple to user

TuomasBorman:
This naming is related to issue/PR that is open in mia. Wait until we have fixed that.

TuomasBorman:
https://github.com/microbiome/mia/pull/623

tests/testthat/test-2plotAbundance.R
```diff
@@ -74,4 +74,11 @@ test_that("plot abundance", {
                           order.col.by = "SampleType")
     expect_true(is.list(plot))
     expect_s3_class(plot[[1]],"ggplot")
+    #
+    rowData(x)$Salame <- sample(letters[1:5], nrow(x), replace=TRUE)
+    plot <- plotAbundance(x, assay.type="counts", rank = "Salame",
```
TuomasBorman: rank --> group

R/plotAbundance.R
```diff
@@ -187,12 +190,13 @@ setMethod("plotAbundance", signature = c("SummarizedExperiment"),
...
-            stop("'rank' must be an non empty single character value or NULL.",
+        if(!.is_non_empty_string(group) && !is.null(group)){
+            stop("'group' must be an non empty single character value or NULL.",
                 call. = FALSE)
         }
-        if(!is.null(rank)){
-            .check_taxonomic_rank(rank, x)
+        if(!is.null(group) && !(group %in% colnames(rowData(x)))){
+                stop("'group' must be a column from rowData .",
+                     call. = FALSE)
```
TuomasBorman: Check indentation. Run BiocCheck and fix everything related to these modified lines

R/plotAbundance.R
```diff
@@ -13,8 +13,10 @@
 #'   \code{\link[SummarizedExperiment:SummarizedExperiment-class]{SummarizedExperiment}}
 #'   object.
 #'
-#' @param rank \code{Character scalar}. Defines the taxonomic rank to
-#'   use. Must be a value of \code{taxonomyRanks(x)}. (Default: \code{NULL})
+#' @param group \code{Character scalar}. Defines a group to use. Must be a value 
+#' of \code{colnames(rowData(x))}. (Default: \code{NULL})
```
TuomasBorman: This must say explicitly that this controls agglomeration.

TuomasBorman:
We decided to prefer that user does (relative) transformation outside the function. That is why as.relative was moved to "hidden" parameter. This means that also agglomeration should be done outside the function. --> I moved rank/group parameter as hidden parameter.

TuomasBorman (approved):

PR #149, opened 2024-09-16
#639- update plotCCA/RDA

0xMuluh:


TuomasBorman (changes requested):
Looks good! However, this is overlapping with larger PR https://github.com/microbiome/miaViz/pull/154

PR #152, opened 2024-10-06
plotLoadings, absolute scale

TuomasBorman:
Now loadings can be visualized with barplot also in absolute scale. That is the default choice.

The reason is that original scale takes too much space since values can be negative and positive. Moreover, the effect size cannot be easily compared between positive and negative values (that is the most important info)

image

In absolute scale, there the comparison is easier. +/- shows whether the value is positive or negative

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: miaViz
 Title: Microbiome Analysis Plotting and Visualization
-Version: 1.13.11
+Version: 1.13.12
 Authors@R: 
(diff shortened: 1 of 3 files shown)


antagomir:
How about distinguishing the positive vs. negative with a color instead?

TuomasBorman:
With colors, it seems to be easier to see +/-

image

antagomir:
Yes. You also sometimes see this done so in literatures.

Just thinking if the default should be lines instead of bars ("[lolliplot plot](https://r-graph-gallery.com/lollipop-plot.html)")?

PR #153, opened 2024-10-07
plotLoadings modifications

TuomasBorman:
Now user can also show the sign with colors. This is the default

image

https://github.com/microbiome/miaViz/pull/152


antagomir:
How about adding "lolliplot plot" option for the layout?

TuomasBorman:
Something like this:

image

antagomir:
Yes, for instance. 

Or, just the lines (with colors, without points). Somehow I like this appearance visually a bit more than bars.

TuomasBorman:
Different plots. Note that adding sign and removing color is hidden options (they are for us to test). By default, if the scale is absolute, color is added. Otherwise the color is not added.

image
image
image
image
image

TuomasBorman:
These are default options for barplot

DEFAULT
image


If absolute.scale is set to FALSE
image

antagomir:
Seems good

PR #156, opened 2024-10-09
Improve plotAbundance

### LINKED ISSUE #132: plotAbundance improvements
1.

When sample names are plotted, one cannot read them as they are over each other
```
library(miaViz)
data("GlobalPatterns")

tse <- GlobalPatterns
plotAbundance(tse, rank = "Phylum", add_x_text = TRUE)
```
image
Some other functions seem to have `angle_x_text` parameter, but `plotAbundance` does not have option to rotate text.

Also, we could consider if sample names could be specified from `colData(tse)`. For example, paired samples must have unique names currently, but better option would be to allow shared names so that one can easily see which samples are drawn from same patient.

2.

I user wants to compare abundances between groups or if samples are paired for instance, our solution might be suboptimal.

```
library(patchwork)
library(miaViz)
data("GlobalPatterns")

tse <- GlobalPatterns
p <- plotAbundance(tse, rank = "Phylum", features = "SampleType")
wrap_plots(p, ncol = 1,  heights = c(0.95,0.05))
```
image
It might be hard to read the plot when there are multiple groups (space between groups might help).

Another option would be to plot abundances as shown [here in figure 1b](https://www.nature.com/articles/s41522-020-0130-5/figures/1)

TuomasBorman:
This PR implements the suggestions made in here: https://github.com/microbiome/miaViz/issues/132

The whole function is now restructured and improved. Even though the function was working, it was too complex to maintain I think. It was not easy to add the suggested things without refactoring the code. That is why the function is now mostly coded again, I think it is much simpler. Moreover, it has many additional capabilities that were not before which makes the function now rather flexible.

Main changes

1. The code is now simpler. However, all old functionalities are still fully supported.
2. User can choose manually the order of columns and rows.
3. Facets are supported now also for columns
4. Support for paired samples.

Because this has many new features and the Bioconductor release is coming soon, I think it is the wisest to **keep this PR open and merge this to the next Bioconductor development version**. I can also test this in one of my projects before merging.

Here are few examples on the usage
```

library(miaViz)
library(patchwork)
data(enterotype, package="mia")
tse <- enterotype
# Apply relative transformation
tse <- agglomerateByPrevalence(tse, prevalence = 0.98)
tse <- tse[ , complete.cases(colData(tse))]
```
```
# Plot with annotation bar
p <- plotAbundance(tse, col.var = "Gender", assay.type = "relabundance")
wrap_plots(p, ncol = 1, heights = c(0.95, 0.05))
```
image
```
# Plot with annotation bar and order
p <- plotAbundance(tse, col.var = "Gender", order.col.by = "Gender", assay.type = "relabundance")
wrap_plots(p, ncol = 1, heights = c(0.95, 0.05))
```
image
```
# Plot with facets for sample metadata variable
p <- plotAbundance(tse, col.var = "Gender", facet.cols = TRUE, scales = "free_x", assay.type = "relabundance")
p
```
image
```
# Plot with facets for multiple sample metadata variables (this requires an additional package that is not dependency)
p <- plotAbundance(tse, col.var = c("Nationality", "Gender"), facet.cols = TRUE, scales = "free_x", assay.type = "relabundance")
p
```
image
```
# Add own taxa order
order <- rownames(tse)
order <- c(order[!order%in%c("Bacteroides")], "Bacteroides")
p <- plotAbundance(tse, row.levels = order, assay.type = "relabundance")
p
```
image
```
# Prepare data
data("Tengeler2020")
tse <- Tengeler2020
tse <- agglomerateByPrevalence(tse, prevalence = 0.70)
tse <- tse[, 1:12]
colData(tse)[["patient"]] <- rep(paste0("patient", seq_len(4)), ncol(tse) / 4)
colData(tse)[["time_point"]] <- rep(paste0("time", seq_len(20)), each = 4)[1:12]

# Drop patient 1 from 1st time point, patient 2 from 2nd time point and patient 3 and 4 from time point 3
tse <- tse[ , c(2:5, 7:10)]
```
```
# Create a plot with paired data for comparison
p <- plotAbundance(
    tse, col.var = "time_point", order.col.by = "patient", as.relative = TRUE,
    paired = TRUE, facet.cols = TRUE, ncol = 1, scales = "free_x")
p
```
image


antagomir:
very nice!

R/plotAbundance.R
```diff
@@ -223,7 +223,8 @@ setMethod("plotAbundance", signature = c("SummarizedExperiment"), function(
             paste0(not_allowed, collapse = "', '"), "'", call. = FALSE)
     }
     # Leave order.col.by out of the comparison, if it specifies a feature
-    order.col.by <- if(order.col.by %in% colnames(colData(x))) order.col.by else
+    order.col.by <- if( !is.null(order.col.by) && order.col.by %in% 
+                       colnames(colData(x))) order.col.by else
```
TuomasBorman: Fix the indentation

R/plotAbundance.R
```diff
@@ -237,7 +237,7 @@ setMethod("plotAbundance", signature = c("SummarizedExperiment"), function(
 # agglomeration and transformation. The outut is a single tibble with all the
 # whole dataset for plotting. 
 .get_abundance_data <- function(
-        x, assay.type, layout, group = rank, rank = NULL,
+        x, assay.type, layout = "bar", group = rank, rank = NULL,
```
TuomasBorman: No need for this since the layout is already declared in line 174

TuomasBorman: @Daenarys8

0xMuluh: test fails  `.get_abundance_data()`  or I could just pass it directly in the test.

TuomasBorman: You can pass directly from the test. It might be easier to see from the interval function which parameters are not yet declared i.e. they are "hidden parameters"

0xMuluh: okay. it's working now

PR #160, opened 2024-10-21
add plot method for eigenvalues

### LINKED ISSUE #159: Plot eigenvalues
Single-cell toolbox and mia do not currently have function for visualizing eigenvalues. Usually, that it done with screeplot that shows how the eigenvalues decreases.

1.

The function name could be `plotScree` or `plotEigenvalues`

2.

We could have methods for `SingleCellExperiment`, and `vector`.

SingleCellExperiment method gets the value from `reducedDim(x, dimred)`, from the correct attribute like functions `plotRDA()` and `plotLoadings()`. It creates a vector from the data with component names in `names()` and call the vector function.

Note that not all functions store the eigenvalues to same place / under same attribute name.

As usually eigenvalues are returned as a vector, we could have a method for vector also. That function could just create a data.frame from the vector. The data.frame could have the eigenvalues in one column and the names of vector in another. 

3.

Eigenvalues are commonly visualized with combination of barplot, point plot, line plot and labels. We could have them all but user could control them with `show.*`. These could be hidden parameters to make the methods less complex.

4.

Moreover, we could have also a cumulative explained variance plot that shows how much variance is explained by the first i component. That could also be a hidden option. Maybe `cumulative = FALSE`. In theory, that just needs one extra step that calculates the cumulative proportion.

image

0xMuluh:
issue: #159


TuomasBorman:
Commenting wirh phone so little bit hard, but:

1.

Eventhough SCE method, in description of x say that the obkect must be TreeSE (not SCE). Some users might not knoe that TreeSE is actually SCE and this might be confusing

2.

Structure the function so that. One internal functiona prepares the data and another internal function plots the data (.scree_plotter).

3.

Put these show.* parameters as a hidden parameters of .scree_plotter

0xMuluh:
i have added update to consider actual variance explained(`VarExplained`) by each PCi in `runPCA` as it does not return eigenvalues. What other `reducedDim` are to be supported?

antagomir:
You could also support MDS (aka PCoA). For other ordinations this is not well defined I think.

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+        # Call the vector method
+        plotScree(as.numeric(eig), names(eig), cumulative = cumulative, ...)
+    }
+)
+
+#' @rdname plotScree
+#' @export
+setMethod("plotScree", signature = c(x = "vector"),
+    function(x, names = NULL, cumulative = FALSE, ...) {
+        # Ensure 'x' is numeric
```
TuomasBorman: Wondering if this is better approach, so we can remove `names` argument

```
eigenvalues <- c(2, 3, 2)
names(eigenvalues) <- c("PC1", "PC2", "PC3")
plotScree(eigenvalues)
```

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+        if( any(ind) ){
+            # Add explained variance
+            eig <- attributes(reduced_dim)[ind][[1]]
+        } else{
+            stop("No eigenvalues found in the specified reducedDim.", 
+                 call. = FALSE)
+        }
+      
+        # Call the vector method
+        plotScree(as.numeric(eig), names(eig), cumulative = cumulative, ...)
```
TuomasBorman: End functions with return()

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+setGeneric("plotScree", signature = c("x"),
+    function(x, ...) standardGeneric("plotScree"))
+
+#' @rdname plotScree
+#' @export
+setMethod("plotScree", signature = c(x = "SingleCellExperiment"),
+    function(x, dimred, cumulative = FALSE, ...) {
+        # Check if dimred exists
+        if (!dimred %in% reducedDimNames(x)) {
+            stop("'dimred' must specify a valid reducedDim.", call. = FALSE)
```
TuomasBorman: We could also support index (that could be supported but not described in the documentation --> simpler when less options)

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+        # Extract eigenvalues
+        # Check if data is available
+        ind <- names(attributes(reduced_dim)) %in% c("eig", "varExplained")
+        if( any(ind) ){
+            # Add explained variance
+            eig <- attributes(reduced_dim)[ind][[1]]
+        } else{
+            stop("No eigenvalues found in the specified reducedDim.", 
+                 call. = FALSE)
+        }
```
TuomasBorman:
```
if( !any(ind) ){
   stop()
}
eig <- attributes(reduced_dim)[ind][[1]]
```

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+            vjust = -0.5)
+    }
+    
+    # Customize appearance
+    p <- p + theme_minimal() +
+        labs(x = "Component", 
+             y = if (cumulative) "Cumulative Proportion of Variance" 
+             else "Eigenvalue",
+             title = if (cumulative) "Cumulative Explained Variance" 
+             else "Scree Plot")
```
TuomasBorman: Do not add title

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+    }
+    if (show.labels) {
+        p <- p + geom_text(aes(label = round(if (cumulative) 
+            CumulativeProportion else Eigenvalue, 2)), 
+            vjust = -0.5)
+    }
+    
+    # Customize appearance
+    p <- p + theme_minimal() +
+        labs(x = "Component",
```
TuomasBorman: X axis title could be PC so it is more general

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+        .scree_plotter(plot_data, cumulative = cumulative, ...)
+    }
+)
+
+.prepare_data <- function(x, names = NULL, cumulative = FALSE) {
+    df <- data.frame(
+        Component = if (!is.null(names)) names else seq_along(x),
+        Eigenvalue = x
+    )
+
```
TuomasBorman: Remove extra line breaks, this is so short function so they do not improve readability

R/plotScree.R
```diff
@@ -0,0 +1,169 @@
...
+.prepare_data <- function(x, names = NULL, cumulative = FALSE) {
+    df <- data.frame(
+        Component = if (!is.null(names)) names else seq_along(x),
+        Eigenvalue = x
+    )
+    
+    # Calculate cumulative proportion if needed
+    if (cumulative) {
+        df$CumulativeProportion <- cumsum(df$Eigenvalue) / sum(df$Eigenvalue)
```
TuomasBorman: What if there are NAs?

TuomasBorman:
> You could also support MDS (aka PCoA). For other ordinations this is not well defined I think.

Also RDA is based on eigenvalues-eigenvectors

R/plotScree.R
0xMuluh: issued warning and removed NAs.

R/plotScree.R
0xMuluh: ACK

R/plotScree.R
0xMuluh: I'm not sure I understand this. Do you mean subsetting the vector based on indices for example `indices = c(4,5)`?

TuomasBorman: Sorry about unclear message, I meant `reducedDim(tse, 1)`, but I think you did it correctly

PR #188, opened 2025-05-02
Additional offset methods for plotBoxplot

TuomasBorman:
Add additional point offset methods for plotBoxplot. Npw quasirandom is used as default.


TuomasBorman:
TODO:

Consider the default offset function. If quasirandom is the default choice, vipor package must be added as dependency (currently it is "soft-dependency")

TuomasBorman:
quasirandom requires complete dataset. That is why we stick "jitter" as default choice.

PR #189, opened 2025-05-03
plotLoadings, sort

### LINKED ISSUE #183: plotLoadings feature sort
It could be handy if the features can be sorted in [plotLoadings](https://microbiome.github.io/miaViz/reference/plotLoadings.html). For instance the heatmaps (at the end of examples) would look more clear if the features were sorted by e.g. hclust / neatsort, or according to PC1 for instance.

TuomasBorman:
https://github.com/microbiome/miaViz/issues/183


TuomasBorman:
See the discussion in OMA: https://github.com/microbiome/OMA/issues/146

It might be that getNeatOrder() will be removed and replaced by utilization of seriation package. In that case, this PR must be modified to use seriation package as well. That is rather small change as the functionality for sorting the loadings is already here but must be taken into account.

We could also consider using seriation package here in any case.

PR #193, opened 2025-05-19
p-values to boxplot

TuomasBorman:
This PR implements the p value option for boxplots. The code is still draft, but should work for 2 group comparison.

Things to do:
1. Generalize this idea to more than 2 groups comparisons
2. Update documentation
3. Update testing

To note: this should work for when user specifies `add.significance=TRUE` (the method calculates p-values) and when user specifies the p-values in a table with `p.value`.

Could you @Daenarys8 be able to implement these. I have very limited time this week.

Messy script that you can use for testing:

```
devtools::load_all()

data("Tengeler2020")
tse <- Tengeler2020

tse <- addAlpha(tse)
plotBoxplot(
    tse, col.var = "shannon_diversity",
    x = "patient_status", fill.by = "patient_status", point.shape = 19,  add.significance = TRUE
)

plotBoxplot(
    tse, assay.type = "relabundance", fill.by = "patient_status",
    features = rownames(tse[1:5, ]), x = "rownames", dodge.width = 1, add.significance = TRUE,
    add.proportion = TRUE, threshold = 0, add.threshold = TRUE)

# Add manually calculated p values to plot
df <- meltSE(tse, add.col = TRUE, row.name = "rownames")
pvals <- df |>
    group_by(rownames, cohort) |>
    rstatix::t_test(counts ~ patient_status) |>
    ungroup() |>
    as.data.frame()
# rstatix adds metadata information. Thsi metadata is not obviously available from other methods used
attributes(pvals)[["args"]] <- NULL
pvals <- pvals[, c("rownames", "cohort", "group1", "group2", "p")]


features <- rownames(tse[1:5, ])
plotBoxplot(
    tse, assay.type = "relabundance", fill.by = "patient_status",
    facet.by = "cohort", x = "rownames", features = rownames(tse[1:5, ]),
    p.value = pvals
    )

plotBoxplot(
    tse, assay.type = "relabundance",
    x = "patient_status", fill.by = "patient_status",
    features = rownames(tse)[1:6], facet.by = "rownames",
    add.proportion = TRUE, threshold = 0, add.threshold = TRUE,, add.significance = TRUE, mark.significance = TRUE,
    scales = "free_y")


plotBoxplot(
    tse, assay.type = "relabundance",
    x = "patient_status", fill.by = "patient_status",
    features = rownames(tse), facet.by = "rownames",, add.significance = TRUE,
    add.proportion = TRUE, threshold = 0, add.threshold = TRUE,
    scales = "free_y")

plotBoxplot(
    tse, assay.type = "relabundance",
    x = "patient_status", add.significance = TRUE,
    add.proportion = TRUE, threshold = 0, add.threshold = TRUE)
```


R/plotBoxplot.R
```diff
@@ -359,6 +363,27 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
...
+    }
+    cols <- c("group1", "group2")
+    p_cols <- c("p", "p.adj", "p.adj.signif")
+    if( is.data.frame(p.value) && (
+            !all(cols %in% colnames(p.value)) ||
+            sum(p_cols %in% colnames(p.value)) == 0L)
+        ){
+        stop("'p.value' must have columns '", paste0(cols, collapse = "', '"),
+            "' and one of the following columns: '",
+            paste0(p_cols, collapse = "', '"), "'", call. = FALSE)
```
TuomasBorman, line 527: What if `facet.by`, `group.by` and `fill.by` are specified, but p values do not have the same grouping?

For instance, if p values includes only one p-value. And not multiple groups that are visualized.

Probably does not work correctly

TuomasBorman: The input check should be improved, it is hard for user to know what to do (what is the correct format)

0xMuluh:
Thanks Yes. I can add to the PR on diff.by default variable.  Or better still push them in different prs.

R/plotBoxplot.R
```diff
@@ -442,6 +483,39 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
...
+        stop("Groups in p.value[['", x, "']] ",
+            "must match with 'x'.", call. = FALSE)
+    }
+    # If facetting is specified in the plot, the same facet variable should be
+    # found from p values.
+    if( !is.null(facet.by) && !is.null(p.value) &&
+            !all(p.value[[facet.by]] %in% df[[facet.by]]) ){
+        stop("Groups in p.value[['", facet.by, "']] ",
+            "must match with 'facet.by'.", call. = FALSE)
+    }
```
TuomasBorman: These could be moved to input check. The p value table calculated by the method should be ok.

R/plotBoxplot.R
```diff
@@ -480,7 +555,7 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
     # first instance. Otherwise it would be time point 1 -> time point 2,
     # which is NA.
     df <- df |>
-        arrange(desc(across(all_of(c(pair.by, x)))))
+        arrange(desc(across(all_of(c("difference", pair.by, x)))))
```
TuomasBorman, line 677: This was added to ensure that the line between paired samples was colored correctly.

R/plotBoxplot.R
```diff
@@ -796,7 +1004,7 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
 
 # This function adds points to plot
 .add_points_layer <- function(
-        p, df, point.alpha = 0.65, point.size = 2, point.shape = 21,
+        p, df, point.alpha = 0.65, point.size = 2, point.shape = 19L,
```
TuomasBorman, line 1232: Default point shape is changed. This does not have fill, so it won't be colored with `fill.by` (like boxplot does)

TuomasBorman:
> Thanks Yes. I can add to the PR on diff.by default variable. Or better still push them in different prs.

The `diff.by` is already defined by `group.by`/`fill.by`/`x`. They already define the grouping done in plotting, so no additional `*.by` parameters needed. The comparisons should follow the same comparisons that are visualized. (if I understood correctly the meaning of the parameter)

0xMuluh:
> > Thanks Yes. I can add to the PR on diff.by default variable. Or better still push them in different prs.
> 
> The `diff.by` is already defined by `group.by`/`fill.by`/`x`. They already define the grouping done in plotting, so no additional `*.by` parameters needed. The comparisons should follow the same comparisons that are visualized. (if I understood correctly the meaning of the parameter)

if i understood correctly, it seems the suggestion was to have param where we could pass a precalculated delta in `colData() `to be used as slopes for the change. Since currently we calculate the delta with `.calculate_paired_difference()` internally. fill.by/ group.by/ x are some sort of factored variables from what i have seen

TuomasBorman:
Ahh, yes. That is different thing. However, do we need `diff.by` if the difference is always defined based on difference between point A and point B, i.e., the` .calculate_paired_difference()` should always work? Moreover, that is easy to compute.

Moreover, if user wants to visualize taxa in a boxplot, where the delta goes? What is the format of delta column (it should  probably be a list if taxa are plotted? This goes easily rather complicated for ordinary user.

The suggestion was to add p-values to plot as that functionality is not yet available. However, in this case we have to allow user-defined p-values as they are often generated from specific, external methods. The problem of the format still arises, as this visualizing method should work despite on whether user is visualizing colData, rowData or assay (single column with a vector of p-values is not sufficient)

antagomir:
For both delta and p-values, it would be very good if the users can also provide/modify them manually if needed (can be through rowData/colData).

TuomasBorman:
My point was that `plotBoxplot()` is designed to be as flexible as possible, allowing users to visualize numeric variables from `rowData`, `colData`, or `assays`.

In simple cases, differences between observations can be directly stored in `colData` or `rowData`. For example, when plotting a variable from colData:

```
library(miaViz)

library(microbiomeDataSets)
mae <- microbiomeDataSets::peerj32()
tse <- getWithColData(mae, 1)
tse[["time_point"]] <- as.character(tse[["time"]])

tse <- addAlpha(tse, index = "shannon")
# Create a plot showing chance between time points in abundance of
# Akkermansia
plotBoxplot(
     tse, x = "time_point", col.var = "shannon", fill.by = "group",
     pair.by = "subject", add.chance = TRUE, scales = "free"
)
```
image

In this case, the difference (e.g. change over time) can be stored as a single vector in `colData`.

However, it becomes more complex when visualizing per-taxon differences between sample pairs. For instance this example:
```

# Visualize relative abundance of top features
tse <- tse[getTop(tse, 6), ]
plotBoxplot(
    tse, x = "time_point", assay.type = "relabundance",
    features = rownames(tse), facet.by = "rownames",
    fill.by = "group",
    pair.by = "subject", add.chance = TRUE, scales = "free"
)
```
image


Here, each sample pair should have a vector of differences (one per taxon). Which causes a problem: where and how should these differences be stored? A few options:

a. Store a nested list in a single column of `colData`, where each element represents a vector of differences for that sample pair. This can become complex and harder to manage.

b. Store each taxon's difference as a separate column in `colData`. This may not scale well with a large number of taxa.

So to conclude:

1. Storing difference to `colData`/`rowData` is easy for the simplest cases, but can be rather complex
2. Difference should be always defined based on difference between sample pairs. It is efficient to calculate for limited number of samples and taxa, and could be handled inside the function for all cases.

If there is a real need for supporting user defined difference values, it is ok to add. But I would not sacrifice usability with this (for instance, requiring completely different format for different cases). A single column in a list format in `colData` might be a reasonable compromise.

P-values are different thing, those we must support (currently in this PR, user should define them as a data.frame, however, we can think about the format after the function works otherwise).

antagomir:
Can be skipped for the time being at least, I guess.

TuomasBorman:
The logic is still failing:

```
devtools::load_all()
data("Tengeler2020")
tse <- Tengeler2020
tse <- tse[1:10, ]

plotBoxplot(
    tse, x = "patient_status", assay.type = "counts",
    features = rownames(tse)
)
plotBoxplot(
    tse, x = "rownames", assay.type = "counts",
    features = rownames(tse)[1:5], add.significance = TRUE
)
```

Let us first fix this:

1.
How to calculate p-values inside the function. Different scenarios must be taken into account (facet.by, group.by, fill.by, x) This is easier.

2. 
Harder task is to implement a way for user to input pre-calculated p-values. Let us do that after the task 1 as this is much harder.

0xMuluh:
This recent update now accounts for all edge cases unless i am missing any(unlikely). If you have more practical examples, I can test to confirm.

TuomasBorman:
Sorry, I have been too busy to review this. I will do it as soon as I can. Meanwhile, can you provide an example that shows how to add p-values that were calculated before calling this function? It should work as flexibly as the function is currently working, i.e., working with facets, grouping, ...

I guess the format of p-values should be long data.frame. That should then have column names equal to ones that are used for plotting variables. For example, in this case

`plotBoxPlot(tse, x = "time", y = "rownames", facet.by = "diagnosis")`

the data.frame should have columns "rownames", "time", "diagnosis", "p_value"

R/plotBoxplot.R
antagomir: @Daenarys8 close if resolved?

R/plotAbundance.R
```diff
@@ -603,7 +603,7 @@ setMethod("plotAbundance", signature = c("SummarizedExperiment"), function(
             ncol = ncol,
             scales = scales)
     }
-    # This is also for facetting based on sample metadata, however, this allows
```
TuomasBorman, line 606: To make review easier (and code maintenance), please modify lines that are rlated to this PR. Not big deal here, but for the future

R/plotBoxplot.R
```diff
@@ -326,22 +459,41 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
...
+        
+        if (length(unique(sample_counts$n_samples)) > 1) {
+            stop(
+                "When 'pair.by' is specified, all subjects must have the ",
+                "same number of samples.\nRemove 'pair.by' or filter your ",
+                "data to include only subjects with balanced repeats.",
+                call. = FALSE
+            )
+        }
+    }
```
TuomasBorman: This is not necessary requirement

R/plotBoxplot.R
```diff
@@ -923,18 +921,28 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
...
+                    p.adjust.method = "none"
+                )
+                # Add grouping columns back - assuming grouping_vars are unique 
+                # per group
+                bind_cols(
+                    res,
+                    df_group %>% select(all_of(grouping_vars)) %>% distinct()
+                )
+            }, error = function(e) NULL)
+        }) |>
```
TuomasBorman: What are these changes? What are the improvements compared to old one?

R/plotBoxplot.R
```diff
@@ -947,13 +955,24 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
...
+            attributes(pvals)[["args"]][["data"]] <- temp
+        } else {
+            # Use 'FeatureID' for subsetting
+            pvals <- pvals[
+                pvals$group1 %in% features & pvals$group2 %in% features,
+                , drop = FALSE]
+            
+            temp <- attributes(pvals)[["args"]][["data"]]
+            temp <- temp[temp[["FeatureID"]] %in% features, , drop = FALSE]
+            attributes(pvals)[["args"]][["data"]] <- temp
```
TuomasBorman, line 1020: What is the intention here?

features refer to taxa, but the pvals$group1 should include the sample grouping?

R/plotBoxplot.R
TuomasBorman: tryCatch hinders the possible problems

R/plotBoxplot.R
```diff
@@ -970,17 +970,24 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
...
+            ]
             
             temp <- attributes(pvals)[["args"]][["data"]]
             temp <- temp[temp[["FeatureID"]] %in% features, , drop = FALSE]
             attributes(pvals)[["args"]][["data"]] <- temp
+            
+        } else {
+            # Third case: explicitly do nothing
+            # No filtering applied to pvals or its attributes
         }
```
TuomasBorman, line 1025: else is not needed

R/plotBoxplot.R
```diff
@@ -1030,34 +1133,61 @@ setMethod("plotBoxplot", signature = c(object = "SummarizedExperiment"),
...
+                color = NULL
+            ),
+            alpha = line.alpha,
+            linetype = linetype,
+            linewidth = linewidth,
+            colour = line.colour
+        )
+        args <- args[lengths(args) > 0]
+        p <- p + do.call(geom_path, args)
     }
```
TuomasBorman, line 1319: What are these changes? These are not related to p-values?

R/plotBoxplot.R
TuomasBorman:
```
plotBoxplot(
    tse,
    fill.by = "diagnosis", assay.type = "counts",
    features = rownames(tse),
    p.value = res
)
```

This is not working, but we can restrict the use of p.value to only cases where x axis specified

R/plotBoxplot.R
0xMuluh: Right. However user has to match sample pairs for paired testing, else NAs are introduced and the testing fails.

TuomasBorman:
The plotBoxplot() functions is working pretty nicely and is rather complex already. This p-value plotting adds lots of complexity and I do not want to break the existing function. This is why I have been thinking creating addPValues() or similar function (similar to `ggsignif`). That would probably make the approach much simpler because all these facets, axes, grouping etc are already defined by the main plotting function.

I will propose idea this week. The work in this PR is not lost, same idea can be used but hopefully much simpler manner.

R/plotBoxplot.R
0xMuluh: This is only in the case when user wants to plot given features.  This is in .calculate_significance() to ensure only the relevant pval to be added to the plot moves forward.

R/plotBoxplot.R
0xMuluh: This updated function enables the flexible line segments for add.change. The previous code was suitable for showing differences between two timepoints (such as `baseline-week10`) but is unable to account for 3 or more timpoints (such as `summer-autumn-winter`)

TuomasBorman:
HI!

I just merged this PR https://github.com/microbiome/miaViz/pull/216/ that implements the p-value functionality with minimal changes to the working code.

As this PR contained also other improvements (such as correction of add.change argument name), can you @0xMuluh implement them in a new PR along with unit tests and do some testing?

Issue #197, opened 2025-07-14
plotRDA fixes


TuomasBorman:
1.

`plotRDA()` does not find vectos even though they are in data. Fix `plotRDA()` so that it finds the interaction results.

```
library(mia)

data("Tengeler2020")

tse <- Tengeler2020

colData(tse)
tse <- addRDA(tse, formula = x ~ patient_status * cohort, assay.type = "counts")

library(miaViz)

plotRDA(tse, "RDA")
```

2.

If there is only one variable in formula, plotRDA() seems to fail when coloring based on that variable.

```
tse <- addRDA(tse, formula = x ~ patient_status, assay.type = "counts")
plotRDA(tse, "RDA", colour.by = "patient_status")

```

> Error in `scale_y_continuous()`:
> ! Discrete values supplied to
>   continuous scale.
> ℹ Example values: "ADHD", "ADHD", "ADHD",
>   "ADHD", and "ADHD"

TuomasBorman:
3. 
```
library(miaViz)
data("Tengeler2020")

tse <- Tengeler2020

tse <- addRDA(tse, formula = x ~ patient_status, assay.type = "counts")

plotRDA(tse, "RDA")
```

The results have only one constrained axis. However, the plot says incorrectly "RDA1" and "RDA2". The "RDA2" should be "MDS" instead.

0xMuluh:
> `plotRDA()` does not find vectos even though they are in data. Fix `plotRDA()` so that it finds the interaction results.
> 
> ```
> tse <- addRDA(tse, formula = x ~ patient_status * cohort, assay.type = "counts")
> 
> library(miaViz)
> 
> plotRDA(tse, "RDA")
> ```
> 
Hi, thanks.
 This is the plot i get, and it seems to have the interaction results, unless I missed your point.

<img width="774" height="566" alt="Image" src="https://github.com/user-attachments/assets/2a2e4a3c-f69e-48a2-b80d-c57c30eb7553" />

TuomasBorman:
> > `plotRDA()` does not find vectos even though they are in data. Fix `plotRDA()` so that it finds the interaction results.
> > ```
> > tse <- addRDA(tse, formula = x ~ patient_status * cohort, assay.type = "counts")
> > 
> > library(miaViz)
> > 
> > plotRDA(tse, "RDA")
> > ```
> 
> Hi, thanks. This is the plot i get, and it seems to have the interaction results, unless I missed your point.
[...]

PR #200, opened 2025-07-15
#197 plotRDA

0xMuluh:


TuomasBorman:
1.

```
library(miaViz)
data("peerj13075")
tse <- peerj13075

tse <- transformAssay(tse, method = "rclr")
tse <- addRDA(tse, assay.type = "rclr", formula = x ~ Geographical_location + Gender * Age)

plotRDA(tse, "RDA")
```

The vectors should have the following p-values

> 
> attr(,"significance")$permanova
>                       Df  Variance        F Pr(>F) Total variance Explained variance
> Model                  7 25.184890 2.434124  0.007       99.08913         0.25416400
> Geographical_location  2 16.991739 5.747890  0.001       99.08913         0.17147934
> Gender:Age             2  3.554816 1.202507  0.294       99.08913         0.03587493
> Residual              50 73.904241       NA     NA       99.08913         0.74583600

So "Geographical_location" and "GenderMale:AgeElderly" should have p-values and explained variance.

Currently, the interaction results are not included.

As the interaction term is just combining two variables, they are constructed always with same format "Gender:Age" (variable1:variable2). You can use that information here: https://github.com/microbiome/miaViz/blob/678c6907caa775dca5f40ccd8cfe22763e462512/R/plotCCA.R#L425

Note that there is no limit on number of terms (one could specify also  `formula = x ~ Geographical_location * Gender * Age`)

TuomasBorman:
2.

Now the names axes names are RDA 1 and MDS. In the results (`reducedDim(tse, "RDA")`) they are as follows:

> dbRDA1         MDS1       MDS2 

So the function overwrites these somewhere. Remove this overwriting, use the names directly from the results.

In this plot, the names should be dbRDA1 and MDS1 (only one constrained axis, and the y-axis is first unconstrained MDS axis).

0xMuluh:
> So the function overwrites these somewhere. Remove this overwriting, use the names directly from the results.
> 

Right. Thanks. This behavior is defined by `plotReducedDim()`
See [overwrite](https://github.com/alanocallaghan/scater/blob/68425da8d6a84031fcf2e2400f68649b3952fd4c/R/plotReducedDim.R#L178)

TuomasBorman:
If I remember, you can pass your own labels (or it might be easiest to modify in miaViz)

0xMuluh:
> If I remember, you can pass your own labels (or it might be easiest to modify in miaViz)

This support is not yet in scater. The latter will do.

0xMuluh:
No need to subset the object if there is case where unconstrained axis will be plotted.

TuomasBorman (approved):
Thanks!

PR #203, opened 2025-07-17
use release in gha runners

0xMuluh:


TuomasBorman:
Thanks! This should fix the problem: https://github.com/markrobinsonuzh/TreeSummarizedExperiment/issues/88

PR #207, opened 2025-10-01
update unit tests

0xMuluh:


TuomasBorman:
Looks good, let's wait to next week if ggtree will be fixed. If not, then we have to think something else as the release  is approaching (October 24).

TuomasBorman (approved):

PR #214, opened 2026-01-05
Fix fill parameter handling in plotHistogram

### LINKED ISSUE #195: fill does not work in plotHistogram
https://github.com/microbiome/miaViz/blob/678c6907caa775dca5f40ccd8cfe22763e462512/R/plotHistogram.R#L302

`fill` parameter does not work, i.e., it does not control the `fill` color. 

This line is incorrect:

https://github.com/microbiome/miaViz/blob/678c6907caa775dca5f40ccd8cfe22763e462512/R/plotHistogram.R#L317

If `fill.by` is not specified, `fill` should not be applied inside mapping. Instead, the line should  be something like this:

`fill = if(!is.null(attributes(df)[["fill.by"]])) .data[[attributes(df)[["fill.by"]]]]`

(i.e. NULL if `fill.by` is not specified)

and this line

https://github.com/microbiome/miaViz/blob/678c6907caa775dca5f40ccd8cfe22763e462512/R/plotHistogram.R#L324

should be something like this

`fill = if(is.null(attributes(df)[["fill.by"]])) fill`

(i.e., `fill` if `fill.by` is not specified)

0xMuluh:
ping #195


R/plotHistogram.R
```diff
@@ -347,16 +323,31 @@ setMethod("plotBarplot", signature = c(x = "SummarizedExperiment"),
         position = ifelse(
             !is.null(attributes(df)[["fill.by"]]), "dodge2", "identity"),
         ...){
+    .plot_hist_bar(df, geom_bar, TRUE, color, fill, alpha, scales,
+        position, ...)
+}
```
TuomasBorman, line 328: End every function with return()

R/plotHistogram.R
```diff
@@ -347,16 +323,31 @@ setMethod("plotBarplot", signature = c(x = "SummarizedExperiment"),
...
+# This function creates a histogram or bar plot from a data.frame.
+.plot_hist_bar <- function(df, geom_fun, use_position, color, fill,
+        alpha, scales, position, ...){
     # To disable "no visible binding for global variable" message in cmdcheck
     value <- facet_by <- NULL
+    fill_by <- attributes(df)[["fill.by"]]
+    mapping <- aes(x = value)
+    if (!is.null(fill_by)) {
+        mapping <- aes(x = value, fill = .data[[fill_by]])
+    }
```
TuomasBorman, line 339: Use the same formatting as in rest of the package

R/plotHistogram.R
```diff
@@ -303,41 +305,15 @@ setMethod("plotBarplot", signature = c(x = "SummarizedExperiment"),
...
-    p <- p + labs(x = attributes(df)[["x"]])
-    if( !is.null(attributes(df)[["fill.by"]]) ){
-        p <- p + labs(fill = attributes(df)[["fill.by"]])
-    } else{
-        p <- p + guides(fill = "none")
-    }
-    return(p)
+    geom_fun <- switch(layout, histogram = geom_histogram, density = geom_density)
+    .plot_hist_bar(df, geom_fun, layout == "histogram", color, fill,
+        alpha, scales, position, ...)
```
TuomasBorman, line 316: use_position is redundant. You can overwrite `position <- NULL` if `layout == "density"`

TuomasBorman (changes requested):
The resulting plot looks good!

TuomasBorman (approved):

Issue #107, opened 2024-01-02
plotAbundanceDensity colors


antagomir:
The `plotAbundanceDensity` function currently allows coloring the points by a `colData` variable:

```
library(miaViz)
library(miaTime)
data("hitchip1006", package="miaTime")
tse <- hitchip1006
plotAbundanceDensity(tse)
```

Sometimes we need to map colors of the features (not points) to other external figures that use the same color coding for those features. For this it would be useful if we could also have a custom color for each visualized features (i.e. the rows in the above jitter plot). Something like:

```
plotAbundanceDensity(tse, assay.type = "relabundance", color_features_by=brewer.pal(n=nrow(tse)))
```

Sometimes it would be useful to color

antagomir:
Related to #90

0xMuluh:
Also related: https://github.com/microbiome/OMA/issues/197

TuomasBorman:
The coloring still colors points. My suggestion is to expand the `colour.by` so that it searches the variable from `rowData` if it cannot be found from `colData`. It might be confusing to have many different coloring arguments.

The exact coloring scheme can be then specified outside the function. The  functions should  give the best default options for coloring and changing it is rather advanced stuff.  

I.e.,

```
rowData(tse)[["butyrate_producing"]] <- sample(c(1, 0), nrow(tse), replace = TRUE)
p <- plotAbundanceDensity(tse, colour.by = "butyrate_producing")
p + scale_fill_manual(values = <whatever-user-wants>, guide = "none")
```

PR #215, opened 2026-01-05
Add colour.features parameter to plotAbundanceDensity

0xMuluh:
ping #107


TuomasBorman (changes requested):
colour.by could be extended to search from rowData if the variable is not found from colData. See for instance, plotHistogram

PR #212, opened 2025-10-29
Fix tree loadings

raivo-otus:
Fixed broken links warned by `devtools::check()`.

Adds `GenomeInfoDb` to suggests in DESCRIPTION, to pass tests. 

Applies "best-effort" default options to `plotLoading()`, these require further work as discussed in issue #211

Touched files were autoformatted to R style guides. **If this is unwanted**, I will revert and make new PR.

PR #217, opened 2026-01-14
Refix tree loadings

### LINKED ISSUE #211: Problems with plotLoadings()
Hey,

There appears to be some issue with the function `plotLoadings()`. It does not respond to the parameters `ncomponents` or `n` and fails to color features. 

Testing with another dataset (GlobalPatterns) replicates the same issue of not responding to the parameters `ncomponents` or `n`, but it will correctly color features. 

Below is a reproducible example with SprockettTHData

```r
# Libraries
library(mia)
library(miaViz)
library(microbiomeDataSets)

# Load data
tse <- SprockettTHData()
# Subset
tse <- tse[, tse[["Sample_Type"]] %in% c("Feces")]
tse <- tse[, tse[["Age_Class"]] %in% c("Infant")]

# Agglomerate and apply transformations
altExp(tse, "Phylum") <- agglomerateByRank(tse, "Phylum")
tse <- transformAssay(tse, method = "relabundance", altexp = "Phylum")
tse <- transformAssay(
  tse,
  method = "clr",
  pseudocount = TRUE,
  altexp = "Phylum"
)

# Perform dbRDA
altExp(tse, "Phylum") <- addRDA(
  altExp(tse, "Phylum"),
  assay.type = "clr",
  method = "euclidean",
  formula = x ~ Age_Months + Country,
  na.action = na.exclude
)

# Plot
p_load <- plotLoadings(
  altExp(tse, "Phylum"),
  dimred = "RDA",
  layout = "heatmap",
  ncomponents = 2,
  n = 20,
  add.tree = TRUE
)

p_load
```

raivo-otus:
Resubmit of fixes applied in #212 WITHOUT autoformatting and unrelated fixes to links


DESCRIPTION
```diff
@@ -66,6 +66,7 @@ Suggests:
     bluster,
     circlize,
     ComplexHeatmap,
+    GenomeInfoDb,
```
TuomasBorman: Is this needed?

R/plotLoadings.R
TuomasBorman, line 1: Wasn't there more improvements? Include all functional improvements related to plotloadings that were in the old PR. (The plot looked very nice in that PR if I remember correctly)

TuomasBorman (changes requested):

R/plotLoadings.R
raivo-otus: I'll double back. Likely missed something, have since disabled all autoformating on save to prevent this mess in the future.

DESCRIPTION
raivo-otus: No, strictly for this no. But checks were throwing annoying errors that this fixed. Along with the correct links that where in the previous PR but omitted here.

TuomasBorman: Can you check what these errors were?

The package is not directly related to miaViz, I think. The dependency comes probably from somewhere else, meaning that the dependency errors should be fixed there.

This is for controlling that the dependency list does not expand too much (the list is already too big)

raivo-otus:
Should be the rest of the changes. Still feels like I might have missed something but can't spot. The function still needs some work I think, it throws many warning on invalid object types when plotting..

DESCRIPTION
raivo-otus: Seems it is not needed anymore. Removed.

TuomasBorman (approved):
Thanks!

PR #226, opened 2026-04-13
Basic rarefaction curve based on vegan::rarecurve

vsuksi:


vsuksi:
Is it important to integrate with mia::getAlpha and niter to be able to visualize multiple alpha indices and make it more general? Initial testing suggests that would not be feasible in terms of performance because first we'd rarefy the assay and then compute the diversity indices. Just observed richness is motivated in that it's intuitive and most sensitive for quality control.

The most overlap I found was with using iNEXT::iNEXT() and iNEXT::ggiNEXT() which is ggplot2-based. Perhaps most relevant downside is that extrapolation can not be turned off and extrapolating species richness curves seems questionable for sequencing data without special singleton estimator (Chao et al. 2016), which is not implemented in iNEXT. One option could be to instruct user to do iNEXT::iNEXT with nboot = FALSE and ggiNEXT.

antagomir:
getAlpha integration is not critical; If it is possible utilize existing methods that's even better - you could see what works best, then we get some experience with it and can improve later if it turns out to be necessary?

TuomasBorman (commented):
Seems very nice!

For bigger datasets, the default options in vegan::rarecurve takes too long time. Could this be made faster somehow?

vsuksi:
I doubt that the actual rarefaction curve calculation can be faster as it already uses an analytical formulation (Hurlbert). I would encourage users in example to reduce the `step` argument. A sensible default for `step` could be `max(colSums(assay(x, assay.type))) / 10`.

I guess I'll finalize this contribution by doing tests and odds and ends. Just to be clear, this function uses `vegan::rarecurve()`, so the counts data needs to include singletons. In the absence of singletons, this will give different results (with warning) compared to the empirical approach in mia package rarefaction framework.

PR #233, opened 2026-09-03
Add multimedia example datasets (se_relative, tse_relative)

YihanLiu4023:
## Summary
Adds two example data objects for demonstrations and examples across miaViz:

- `se_relative` — `SummarizedExperiment` with relative abundance data
- `tse_relative` — `TreeSummarizedExperiment` with the same data plus a phylogenetic tree


R/miaViz.R
```diff
@@ -23,26 +23,26 @@ NULL
...
-#' 
+#'
 #' @details
 #' For \code{*_graph} data:
-#' 
-#' 1. \dQuote{Jaccard} distances were calculated via 
+#'
+#' 1. \dQuote{Jaccard} distances were calculated via
 #' \code{calculateDistance(genus, FUN = vegan::vegdist, method = "jaccard",
 #' exprs_values = "relabundance")}, either using transposed assay data or not
```
antagomir, line 34: The "exprs_values" argument is deprecated (and I believe this is said in a deprecation warning when running this function). Use "assay.type" instead.

R/miaViz.R
```diff
@@ -54,3 +54,61 @@ NULL
...
+#' @keywords datasets
+#' @examples
+#' data(se_relative)
+#' data(tse_relative)
+#' @usage data(se_relative)
+"se_relative"
+
+#' @name multimedia-datasets
+#' @usage data(tse_relative)
+"tse_relative"
```
antagomir, line 114: Example data can be loaded from many sources with different names. It is recommended to use a more descriptive name for the data, for instance AuthorNameYear. Then for running examples you can rename the data object generically (easier to run from example code that always uses se/tse) and the assay type is in assay name assayNames(tse) can be "relative" if you wish to set it so during data preparation. The se/tse object may have also other assay types, so it can be misleading the name the entire object by one assay type (e.g. tse_relative).


See e.g.

```
library(mia)
data(GlobalPatterns, package="mia")
tse <- GlobalPatterns
assayNames(tse)
```

R/miaViz.R
```diff
@@ -54,3 +54,61 @@ NULL
...
+#' \code{relative_abundance}, \code{relabundance}, \code{scaled}.
+#'
+#' \strong{tse_relative} — \code{TreeSummarizedExperiment}. Same assays as
+#' \code{se_relative} plus an attached row tree.
+#'
+#' Shared \code{colData} columns include: \code{study_name},
+#' \code{subject_id}, \code{body_site}, \code{study_condition},
+#' \code{disease}, \code{age}, \code{age_category}, \code{gender},
+#' \code{country}, \code{BMI}, \code{visit_number}, \code{disease_subtype},
+#' \code{disease_binary}, \code{dysbiosis}, \code{Time_point},
```
antagomir, line 89: just wondering if "Time_point" could be "time_point" to be consistent with capitalization

antagomir (commented):
Just minor suggestions.

inst/scripts/multimedia_relative_data_prep.R
```diff
@@ -0,0 +1,138 @@
...
+
+tse_relative <- tse
+
+################
+# Save as .rda #
+################
+
+# One object per file, matching the layout of the other miaViz datasets
+# (col_graph.rda, row_graph.rda, ...).
+save(se_relative,  file = "data/se_relative.rda",  compress = "xz")
```
TuomasBorman, line 137: To save space, only TreeSE could be saved. Also its name could be more describing, e.g., "HMP_2019_ibdmdb"

inst/scripts/multimedia_relative_data_prep.R
```diff
@@ -0,0 +1,138 @@
...
+
+#################################
+# Calculate the dysbiosis score #
+#################################
+
+# For each sample i, compute the median distance to all healthy reference
+# samples (excluding sample i itself).
+sample_ids <- tse$SampleID
+
+tse$dysbiosis <- sapply(seq_along(tse$disease_binary), function(i) {
```
TuomasBorman, line 85: Is this "dysbiosis score" common thing? If it is, it could be added to e.g., mia::addAlpha

antagomir: Yes in a way it is. But this should be a separate issue and not postpone merging this one imo (this has taken really long time already). We could make a note there to switch to that when ready,.

antagomir: @microsud has R package [dysbiosisR](https://microsud.github.io/dysbiosisR) but only in Github I think. Maybe it could be integrated with his permission.

TuomasBorman: > Yes in a way it is. But this should be a separate issue and not postpone merging this one imo (this has taken really long time already). We could make a note there to switch to that when ready,.

I agree that this should not be postponed because of that, but as the chapter is using "dysbiosis score", reader probably wonders how to calculate them so this could be good addition

